PR #7, opened 2026-03-14
feat(stop): add working agent stop/interrupt control

VivaLaMikey:
## Summary

- The stop button in the web UI was already rendered but never wired up — `onInterrupt` was not passed from the chat page to `ChatInput`
- No dedicated stop endpoint existed; deleting a session killed the process but also destroyed the session history
- Added `POST /api/sessions/:id/stop` which kills the engine process, clears queued messages, and keeps the session intact
- Interrupted sessions now resolve as clean `idle` state (not `error`) so no spurious error indicators appear

## Changes

- `SessionQueue`: added `clearQueue()`/`clearCancelled()` for pending task cancellation with a cancellation guard in `enqueue`
- `api.ts`: new `POST /api/sessions/:id/stop` endpoint + `clearCancelled` call before new messages to reset stop state
- `manager.ts`: `result.error?.startsWith("Interrupted")` treated as clean stop — status `idle`, no error, no connector reply
- `web/lib/api.ts`: added `stopSession(id)`
- `chat/page.tsx`: added `handleInterrupt` callback, `session:stopped` WS event handler, wired `onInterrupt` to `ChatInput`

## Test plan

- [ ] Send a long-running prompt, click the stop button — streaming stops immediately, loading indicator clears
- [ ] Verify session status is `idle` (not `error`) after stopping
- [ ] Send a new message to the same session after stopping — it runs normally
- [ ] Queue two messages while one runs, click stop — none of the queued messages execute
- [ ] `curl -X POST http://localhost:7777/api/sessions/<id>/stop` returns `{ "status": "stopped" }`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/jimmy/package.json
+++ b/packages/jimmy/package.json
@@ -1,6 +1,6 @@
 {
   "name": "jinn-cli",
-  "version": "0.3.0",
+  "version": "0.3.1",
   "description": "Lightweight AI gateway daemon orchestrating Claude Code and Codex",
   "license": "MIT",
   "type": "module",
--- a/packages/jimmy/src/gateway/api.ts
+++ b/packages/jimmy/src/gateway/api.ts
@@ -225,6 +225,21 @@ export async function handleApiRequest(
       return json(res, { status: "deleted" });
     }
 
+    // POST /api/sessions/:id/stop
+    params = matchRoute("/api/sessions/:id/stop", pathname);
+    if (method === "POST" && params) {
+      const session = getSession(params.id);
+      if (!session) return notFound(res);
+      const engine = context.sessionManager.getEngine(session.engine);
+      if (engine && isInterruptibleEngine(engine) && engine.isAlive(params.id)) {
+        engine.kill(params.id, "Interrupted by user");
+      }
+      context.sessionManager.getQueue().clearQueue(session.sessionKey || session.sourceRef || session.id);
+      updateSession(params.id, { status: "idle", lastActivity: new Date().toISOString(), lastError: null });
+      context.emit("session:stopped", { sessionId: params.id });
+      return json(res, { status: "stopped", sessionId: params.id });
+    }
+
     // POST /api/sessions/bulk-delete
     if (method === "POST" && pathname === "/api/sessions/bulk-delete") {
       const body = JSON.parse(await readBody(req));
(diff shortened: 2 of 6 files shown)


jimmyenglishpro (commented):
## Code Review — PR #7: Agent stop/interrupt control

**Overall: High-value feature, well-scoped.** The stop button is a real UX gap. Implementation is clean but has a few concerns:

### Issues

**1. Fragile interrupt detection via string matching**
```typescript
const wasInterrupted = result.error?.startsWith("Interrupted");
```
This relies on the engine error message starting with "Interrupted". If the engine changes its error format, or if a legitimate error happens to start with "Interrupted", this breaks silently. 

**Suggestion:** Use a structured field instead:
```typescript
// In EngineResult interface:
interrupted?: boolean;
```
Or at minimum, use a constant: `const INTERRUPT_PREFIX = "Interrupted";`

**2. Potential race condition on stop**
The `/stop` endpoint does three things in sequence:
1. `engine.kill()` — async subprocess termination
2. `queue.clearQueue()` — marks session as cancelled
3. `updateSession()` — sets status to idle

But the running engine task's `finally` block in the queue will also call `running.delete()`, `decrementPending()`, and the manager's result handler will try to `updateSession()` again. This could result in:
- Double status update (stop sets idle, then result handler sets idle/error)
- The manager's `replyMessage` might fire before the interrupt is detected

This is likely benign in practice (both paths set idle), but it's a race worth documenting or guarding.

**3. Cancelled set grows unbounded**
`this.cancelled` is a `Set<string>` that grows with each stop. Items are only removed via `clearCancelled()` (called on new message). If sessions are stopped but never receive new messages, the set accumulates indefinitely.

**Suggestion:** Clear the cancelled entry in the queue's `finally` block, or use a `WeakRef`-based approach if keys are objects. Since keys are strings, consider clearing after the queue drains.

**4. Gratuitous comment removal in manager.ts**
The diff removes several useful inline comments (`// Clean up temp MCP config`, `// Track cost`, `// Clear typing indicator`). These comments aid readability — removing them makes the code harder to follow without adding any value. Please keep them.

**5. `onInterrupt` prop**
The `ChatInput` component receives `onInterrupt` but the diff doesn't show how `ChatInput` uses it. Is the stop button rendered there? If so, the `ChatInput` diff is missing from this PR.

### Positives
- Clean API design (`POST /api/sessions/:id/stop`)
- Good WebSocket event integration (`session:stopped`)
- Queue clearing logic is correct — cancelled tasks are skipped via the `cancelled` set
- Properly clears streaming state in the UI on stop

**Verdict: Approve with changes** — fix the string matching, keep the comments, and address the cancelled set cleanup.

PR #8, opened 2026-03-14
feat(discord): add Discord messaging connector

VivaLaMikey:
## Summary

- Adds Discord as a first-class messaging connector, modelled exactly on the existing Slack connector
- Users can DM the bot or message it in a server channel; each conversation becomes a persistent Jinn session
- Supports Discord threads — messages inside a thread continue the same session

## Changes

- New `connectors/discord/index.ts` — full `Connector` interface implementation
- New `connectors/discord/format.ts` — message splitting at Discord's 2000-char limit
- New `connectors/discord/threads.ts` — session key derivation (`discord:dm:<id>`, `discord:thread:<id>`, `discord:<channel>:<msg>`)
- `shared/types.ts` — `DiscordConnectorConfig` interface + `discord?` field in `JinnConfig.connectors`
- `gateway/server.ts` — conditional Discord startup from config
- `settings/page.tsx` — Discord configuration UI (Bot Token, Allow From, Guild ID)

## Configuration

```yaml
connectors:
  discord:
    botToken: "YOUR_BOT_TOKEN"
    allowFrom: []          # empty = allow all; list of Discord user IDs to restrict
    ignoreOldMessagesOnBoot: true
    guildId: ""            # optional, restrict to one server
```

## Test plan

- [ ] Start gateway with `discord.botToken` set — log shows "Discord connector started"
- [ ] DM the bot — a Jinn session appears in the web UI
- [ ] Message the bot in a server channel — separate session created
- [ ] Reply inside a Discord thread — continues the same session
- [ ] `GET /api/status` shows `connectors.discord.status: "ok"`
- [ ] `allowFrom` with a user ID blocks messages from other users

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/jimmy/package.json
+++ b/packages/jimmy/package.json
@@ -1,6 +1,6 @@
 {
   "name": "jinn-cli",
-  "version": "0.3.0",
+  "version": "0.3.1",
   "description": "Lightweight AI gateway daemon orchestrating Claude Code and Codex",
   "license": "MIT",
   "type": "module",
--- a/packages/jimmy/src/gateway/api.ts
+++ b/packages/jimmy/src/gateway/api.ts
@@ -225,6 +225,21 @@ export async function handleApiRequest(
       return json(res, { status: "deleted" });
     }
 
+    // POST /api/sessions/:id/stop
+    params = matchRoute("/api/sessions/:id/stop", pathname);
+    if (method === "POST" && params) {
+      const session = getSession(params.id);
+      if (!session) return notFound(res);
+      const engine = context.sessionManager.getEngine(session.engine);
+      if (engine && isInterruptibleEngine(engine) && engine.isAlive(params.id)) {
+        engine.kill(params.id, "Interrupted by user");
+      }
+      context.sessionManager.getQueue().clearQueue(session.sessionKey || session.sourceRef || session.id);
+      updateSession(params.id, { status: "idle", lastActivity: new Date().toISOString(), lastError: null });
+      context.emit("session:stopped", { sessionId: params.id });
+      return json(res, { status: "stopped", sessionId: params.id });
+    }
+
     // POST /api/sessions/bulk-delete
     if (method === "POST" && pathname === "/api/sessions/bulk-delete") {
       const body = JSON.parse(await readBody(req));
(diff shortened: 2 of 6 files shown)


jimmyenglishpro (commented):
## Code Review — PR #8: Discord connector

**Overall: Clean implementation that follows existing patterns well.** However, there's a critical missing dependency and a security issue.

### Critical

**1. Missing `discord.js` dependency**
`packages/jimmy/package.json` does NOT add `discord.js` as a dependency. The import at the top of `index.ts` will fail at runtime. The diff only shows a version bump (0.3.0 → 0.3.1). This is a build-breaking omission.

**Fix:** Add `"discord.js": "^14.x"` to dependencies.

**2. Path traversal in `downloadAttachment`**
```typescript
const destPath = path.join(destDir, filename);
```
`filename` comes directly from Discord's `att.name` (user-controlled). A malicious filename like `../../etc/cron.d/evil` would write outside the temp directory.

**Fix:**
```typescript
const safeName = path.basename(filename); // Strip directory components
const destPath = path.join(destDir, safeName);
```

### Issues

**3. Duplicate `DiscordConnectorConfig` interface**
Defined in both `connectors/discord/index.ts` (line 21) and `shared/types.ts` (line 230). The types.ts version should be the single source of truth — the connector should import from there, not redefine it.

**4. Session key for root channel messages**
```typescript
return `discord:${message.channel.id}:${message.id}`;
```
Every non-DM, non-thread message creates a brand new session. In Discord, channels are long-lived conversations (unlike Slack threads). Users might expect a persistent session per channel. Consider `discord:channel:${message.channel.id}` as an alternative, or make this configurable.

**5. No parent context injection**
Slack's connector fetches the thread root message and prepends it as context for the engine. Discord threads don't get this treatment — the engine receives the reply without knowing what it's replying to. This could lead to confused responses in thread conversations.

**6. Emoji reactions limited to Unicode**
```typescript
await msg.react(emoji);
```
Discord reactions need emoji objects (with snowflake IDs) for custom/server emojis. The current implementation only works for standard Unicode emojis. Since the gateway typically uses simple emojis like "eyes" and "white_check_mark", this might need emoji name → Unicode mapping.

**7. Format module uses dynamic imports unnecessarily**
```typescript
const { default: fs } = await import("node:fs");
const { default: path } = await import("node:path");
```
These are Node.js built-ins — just use static imports at the top of the file. Dynamic imports add unnecessary async overhead for modules that are always available.

### Positives
- Clean structure mirroring the Slack connector pattern (format.ts, threads.ts, index.ts)
- Good intent selection (Guilds, GuildMessages, MessageContent, DirectMessages, GuildMessageReactions)
- Proper DM/thread/channel handling in session key derivation
- Error handling is consistent with Slack patterns
- Settings UI is clean and follows existing conventions
- Handles attachments

**Verdict: Request changes** — fix the missing dependency, path traversal, and duplicate interface. The rest can be addressed in follow-ups.

PR #44, opened 2026-04-09
fix: skip Slack unfurl events that crash Claude engine

### LINKED ISSUE #43: Slack unfurl events crash Claude engine with empty prompt
## Bug

When a Slack message contains a URL, Slack sends a secondary "unfurl" event ~1 second later with `user=undefined` and `text=""`. The Slack connector treats this as a regular message, routing an empty prompt to the Claude engine, which crashes with:

```
Error: Input must be provided either through stdin or as a prompt argument when using --print
```

## Reproduction

1. Send a message containing a URL via Slack DM to the bot
2. Observe in `gateway.log`:

```
[INFO] [slack] Received message event: user=U04RU1QMRNG text="My message with https://example.com"
[INFO] [slack] Received message event: user=undefined text=""
[ERROR] Claude exited with code 1: Error: Input must be provided either through stdin or as a prompt argument when using --print
```

The pattern is consistent — every message with a URL triggers the crash.

## Root cause

In `src/connectors/slack/index.ts`, the `app.message()` handler checks for `event.bot_id` but does not check for `event.user`. URL unfurl events have no `user` field, so they pass through all guards and reach the engine with an empty prompt.

## Suggested fix

Add a guard after the `bot_id` check:

```typescript
if (!event.user) {
    logger.debug(`[slack] Skipping event with no user (likely URL unfurl)`);
    return;
}
```

## Environment

- jinn-cli v0.9.3
- Slack socket mode
- Claude engine

MarockNRoll:
## Summary

- Skip Slack message events with no `user` field (URL unfurl ghosts) in the Slack connector
- Prevents empty prompts from reaching the Claude engine and crashing with exit code 1

## Problem

When a message contains a URL, Slack fires a secondary unfurl event ~1s later with `user=undefined` and `text=""`. This event passes all existing guards and routes an empty prompt to Claude:

```
[INFO] [slack] Received message event: user=U04RU... text="Message with https://..."
[INFO] [slack] Received message event: user=undefined text=""
[ERROR] Claude exited with code 1: Error: Input must be provided either through stdin or as a prompt argument when using --print
```

## Fix

4-line guard in `packages/jimmy/src/connectors/slack/index.ts`, right after the existing `bot_id` check:

```typescript
if (!(event as any).user) {
    logger.debug(`[slack] Skipping event with no user (likely URL unfurl)`);
    return;
}
```

Fixes #43

## Test plan

- [x] Send a Slack message with a URL → no crash, message processed normally
- [x] Send a Slack message without a URL → still works
- [x] Bot messages still filtered by existing `bot_id` guard

### DIFF
--- a/packages/jimmy/src/connectors/slack/index.ts
+++ b/packages/jimmy/src/connectors/slack/index.ts
@@ -97,6 +97,11 @@ export class SlackConnector implements Connector {
         logger.info(`[slack] Skipping bot message`);
         return;
       }
+      // Skip ghost events from URL unfurls (user=undefined, text="")
+      if (!(event as any).user) {
+        logger.debug(`[slack] Skipping event with no user (likely URL unfurl)`);
+        return;
+      }
       if (!this.handler) {
         logger.info(`[slack] No handler registered, dropping message`);
         return;


hristo2612:
Thanks for catching that! That's a critical finding. Keep up the good work.

MarockNRoll:
Thanks Chris. Great project btw, been using it daily with Slack. Happy to help if anything else comes up.

PR #46, opened 2026-04-12
fix(slack): add app_mention handler for @mentions in channels

lisovet:
## Summary
- Adds an `app_mention` event handler so the bot responds to `@Bot` mentions in channels
- Filters root-level channel messages in `app.message()` so the bot doesn't respond to every message — only thread replies pass through
- DM behavior is unchanged

## Problem
The Slack connector only registered `app.message()`, which doesn't fire for `app_mention` events. Users adding the bot to a channel and @mentioning it got no response.

Additionally, if `message.channels` was subscribed, *every* channel message would trigger a response — not just @mentions.

## Flow after this change
| Context | Handler | Behavior |
|---------|---------|----------|
| `@Bot` in channel | `app_mention` | Starts a new conversation thread |
| Thread reply in channel | `app.message()` | Continues existing conversation |
| DM | `app.message()` | Works as before |

## Test plan
- [ ] Add bot to a channel, send `@Bot hello` — should get a response in a thread
- [ ] Reply in the thread without @mention — should continue the conversation
- [ ] Send a regular message in the channel (no @mention) — should be ignored
- [ ] DM the bot — should work as before

### DIFF
--- a/packages/jimmy/src/connectors/slack/index.ts
+++ b/packages/jimmy/src/connectors/slack/index.ts
@@ -115,6 +115,13 @@ export class SlackConnector implements Connector {
         return;
       }
 
+      // In channels, only process thread replies (continuing a conversation).
+      // Root channel messages are handled by app_mention instead.
+      if ((event as any).channel_type !== "im" && !(event as any).thread_ts) {
+        logger.debug(`[slack] Skipping root channel message (use @mention instead)`);
+        return;
+      }
+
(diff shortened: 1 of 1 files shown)


hristo2612:
Reviewed — handler correctly mirrors the existing `app.message()` flow and the channel-message filter properly preserves DMs (`channel_type === "im"`) and thread replies. Merging.

PR #47, opened 2026-04-12
fix(web): polyfill crypto.randomUUID for non-secure contexts

lisovet:
## Summary
- Adds an inline `crypto.randomUUID` polyfill in the root layout, before any app code runs
- Uses `crypto.getRandomValues()` as the entropy source (available in all contexts)

## Problem
`crypto.randomUUID()` is only available in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) (HTTPS or `localhost`). When accessing the web UI over plain HTTP on a LAN IP or Tailscale hostname (e.g. `http://my-server.tail12345.ts.net:7777`), the chat page crashes with:

```
TypeError: crypto.randomUUID is not a function
```

The error boundary catches it as `[ChatErrorBoundary] crypto.randomUUID is not a function`, resulting in a blank/crashed chat page.

## Test plan
- [ ] Access `/chat` via `http://localhost:7777` — should work (already did)
- [ ] Access `/chat` via `http://<lan-ip>:7777` — should now work instead of crashing
- [ ] Access `/chat` via `http://<tailscale-hostname>:7777` — should now work
- [ ] Access `/chat` via HTTPS — polyfill is a no-op, native `randomUUID` used

### DIFF
--- a/packages/web/src/app/layout.tsx
+++ b/packages/web/src/app/layout.tsx
@@ -17,6 +17,12 @@ export default function RootLayout({ children }: { children: React.ReactNode }) 
   return (
     <html lang="en" suppressHydrationWarning>
       <head>
+        {/* Polyfill crypto.randomUUID for non-secure contexts (plain HTTP over LAN/Tailscale) */}
+        <script
+          dangerouslySetInnerHTML={{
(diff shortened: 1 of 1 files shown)


hristo2612:
Reviewed — clean polyfill, the static template literal has no user-input surface and the native `randomUUID` path is preserved when the secure-context check passes. Merging.

PR #59, opened 2026-05-19
feat(telegram): wire voice/audio messages through bundled STT module

### LINKED ISSUE #54: Telegram voice messages bypass the bundled STT module — empty text reaches Claude and breaks session resume
### Summary

The repo already ships a `stt` module (`packages/jimmy/src/stt/stt.ts`) plus a portal flow to download Whisper models, and a working `stt.enabled / model / languages` config block. However, the Telegram connector (`packages/jimmy/src/connectors/telegram/index.ts`) only forwards `telegramMsg.text` and never invokes STT. When a user sends a voice/audio/video-note message, the connector emits an `IncomingMessage` with `text: ""`, which then crashes Claude session resume with:

```
Error: No deferred tool marker found in the resumed session.
Either the session was not deferred, the marker is stale (tool already ran),
or it exceeds the tail-scan window. Provide a prompt to continue the conversation.
```

### Reproduction

1. Configure a Telegram connector and enable STT in `~/.jinn/config.yaml`:
   ```yaml
   connectors:
     telegram:
       botToken: ...
       allowFrom: [...]
   stt:
     enabled: true
     model: small
     languages: [en, pt]
   ```
2. Use the portal to download the Whisper `small` model (works — file lands in `~/.jinn/models/whisper/ggml-small.bin`).
3. Install runtime deps the STT module shells out to: `whisper-cli` (whisper.cpp) and `ffmpeg` on PATH.
4. Send a voice message to the bot.
5. The connector forwards an empty-text message; gateway logs show:
   ```
   [ERROR] Claude exited with code 1: Error: No deferred tool marker found in the resumed session.
   [WARN]  Dead session detected for <id>: ...
   ```

### Root cause

`packages/jimmy/src/connectors/telegram/index.ts` only reads `telegramMsg.text`. There is no detection of `voice`, `audio`, or `video_note`, no call into the existing `stt/stt.ts` module, no file download from the Telegram Bot API.

### Proposed fix

Wire the existing `transcribe()` function into the Telegram message handler. Two small changes:

**1. `packages/jimmy/src/shared/types.ts`** — extend `TelegramConnectorConfig`:

```ts
export interface TelegramConnectorConfig {
  id?: string;
  employee?: string;
  botToken: string;
  allowFrom?: number[];
  ignoreOldMessagesOnBoot?: boolean;
  /** STT settings (forwarded from top-level `config.stt`) */
  stt?: GatewayConfig["stt"];
}
```

**2. `packages/jimmy/src/gateway/server.ts`** — pass `config.stt` into `TelegramConnector` at each construction site (3 today: the singleton telegram block, plus the two named-instance branches):

```ts
const telegram = new TelegramConnector({
  botToken: config.connectors.telegram.botToken,
  allowFrom: config.connectors.telegram.allowFrom,
  ignoreOldMessagesOnBoot: config.connectors.telegram.ignoreOldMessagesOnBoot,
  stt: config.stt,
});
```

**3. `packages/jimmy/src/connectors/telegram/index.ts`** — detect voice-like messages and transcribe before forwarding:

```ts
import fs from "node:fs";
import os from "node:os";
import path from "node:path";
import {
  transcribe as sttTranscribe,
  resolveLanguages,
  getModelPath,
} from "../../stt/stt.js";

// in constructor:
private readonly sttConfig?: TelegramConnectorConfig["stt"];

constructor(config: TelegramConnectorConfig) {
  // ...existing assignments...
  this.sttConfig = config.stt;
}

// inside the bot.on("message", ...) handler, before building `msg`:
let messageText = telegramMsg.text || telegramMsg.caption || "";

const voiceLike =
  telegramMsg.voice ?? telegramMsg.audio ?? telegramMsg.video_note;

if (voiceLike && this.sttConfig?.enabled) {
  try {
    const model = this.sttConfig.model || "small";
    if (!getModelPath(model)) {
      logger.warn(`[telegram] STT model '${model}' not downloaded — skipping`);
    } else {
      const langs = resolveLanguages(this.sttConfig);
      const language = langs.length === 1 ? langs[0] : "auto";
      logger.info(
        `[telegram] Transcribing voice message (${voiceLike.duration}s, lang=${language})`,
      );
      try { await this.bot.sendChatAction(telegramMsg.chat.id, "typing"); } catch {}

      const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "tg-stt-"));
      try {
        const localPath = await this.bot.downloadFile(voiceLike.file_id, tmpDir);
        const text = await sttTranscribe(localPath, model, language);
        if (text) {
          messageText = messageText ? `${messageText}\n\n${text}` : text;
          logger.info(`[telegram] Transcribed ${text.length} chars`);
        } else {
          logger.warn("[telegram] Transcription returned empty text");
        }
      } finally {
        try { fs.rmSync(tmpDir, { recursive: true, force: true }); } catch {}
      }
    }
  } catch (err) {
    logger.error(
      `[telegram] STT failed: ${err instanceof Error ? err.message : err}`,
    );
    try {
      await this.bot.sendMessage(
        telegramMsg.chat.id,
        "⚠️ Couldn't transcribe your voice message. Please try again or type instead.",
      );
    } catch {}
    return; // don't forward an empty message downstream
  }
}

// then use `messageText` in place of `telegramMsg.text || ""` when constructing `msg`
```

### Related: end-to-end setup from the gateway only

The portal's "Download Whisper" flow is misleading today: it downloads the model, but the user still has to leave the gateway, install `whisper-cli` (build from source) and `ffmpeg` manually, and put them on the gateway's PATH. Without these, transcription fails silently with `ENOENT: whisper-cli` and the user is left wondering why their voice messages don't work.

To make voice messages truly one-click, the gateway should be able to provision the runtime deps the same way it provisions the model. Some options, from simplest to most complete:

1. **Status surfacing.** `GET /api/stt/status` already exists for the model — extend it to report binary presence (`whisperCli: "found" | "missing"`, `ffmpeg: "found" | "missing"`) and surface that in the portal. Today the portal happily says "ready" while the binary is missing.
2. **Setup wizard with copy-paste install commands.** When binaries are missing, the portal shows a per-OS snippet (e.g. `brew install ffmpeg whisper-cpp` on macOS, build steps for Linux). Low engineering cost, still requires the user to run shell commands.
3. **Managed bundled binaries.** Mirror the model-download pattern: the gateway downloads pre-built static binaries for the platform (whisper.cpp ships releases; ffmpeg has reliable static builds at johnvansickle.com/ffmpeg) into a managed directory (e.g. `~/.jinn/bin/`), and the STT module resolves `WHISPER_CLI` / `FFMPEG` from there before falling back to PATH. This makes voice messages work with zero terminal time.

I'd vote for #3 long-term and #1 as an immediate win — the current "model downloaded but binaries missing" state is the worst UX (looks like it should work, doesn't).

For reference, the manual install I had to do on Amazon Linux 2023 to get past the missing binaries:

```bash
# ffmpeg (static build)
curl -sL https://johnvansickle.com/ffmpeg/releases/ffmpeg-release-amd64-static.tar.xz \
  | tar -xJ -C /tmp
sudo cp /tmp/ffmpeg-*-amd64-static/{ffmpeg,ffprobe} /usr/local/bin/

# whisper-cli (build from source)
sudo dnf install -y git cmake gcc-c++ make
git clone --depth=1 https://github.com/ggml-org/whisper.cpp /tmp/whisper.cpp
cd /tmp/whisper.cpp
cmake -B build -DCMAKE_BUILD_TYPE=Release && cmake --build build -j --target whisper-cli
sudo cp build/bin/whisper-cli /usr/local/bin/
sudo cp build/src/libwhisper.so* build/ggml/src/libggml*.so* /usr/local/lib/
sudo ldconfig

# ensure on the gateway's PATH (default PATH lacks /usr/local/bin)
sudo ln -sf /usr/local/bin/{ffmpeg,whisper-cli} /usr/bin/
```

A user who can do all that probably doesn't need a managed gateway in the first place.

### Notes / open questions

- I prototyped this against the published `0.10.0` dist on my own instance and it works end-to-end (Whisper `small`, EN+PT, ~30s CPU per 6s audio on a t3-class box).
- Auto language detection (`-l auto`) is used when `languages` has more than one entry; falls back to the single configured language otherwise. This matches the existing `resolveLanguages` helper.
- On STT failure the connector now drops the message and notifies the user instead of forwarding an empty message — this also fixes the "No deferred tool marker" crash because the engine never receives an empty resume.
- Open question: should other connectors (Slack, Discord, WhatsApp) get the same treatment? Voice-message support is most natural in Telegram and WhatsApp; Slack/Discord use `audio` attachments differently.
- Happy to send a PR if the approach looks right.

### Environment

- `jinn-cli@0.10.0`
- Node 24.15.0
- Amazon Linux 2023
- whisper.cpp (built from source), ffmpeg static 7.0.2

jeffgardnerdev:
Closes #54.

## Problem

The portal's "Download Whisper" feature downloads the model, but the Telegram connector never invokes STT. Voice messages arrive at the engine with `text: ""`, which crashes Claude session resume with the "No deferred tool marker" error.

## Change

Wires `voice`, `audio`, and `video_note` Telegram messages through the existing `stt/stt.ts` module before they reach the engine.

- Pass `stt` config from gateway through to `TelegramConnector` (3 sites in `gateway/server.ts`)
- In `connectors/telegram/index.ts`: detect voice/audio/video_note, send typing action when this message's turn comes, `bot.downloadFile()` into a per-message tempdir, call `sttTranscribe()`, prepend the transcript to `text`
- Multi-language config → `auto`; single-language → use it directly
- Tempdir is cleaned up in `finally` on both success and failure paths

### Empty-text fall-through (the actual #54 bug)

A voice message must never reach the handler with `text: ""`. Four paths now drop with a user-facing Telegram explanation and `return`:

| Condition | User-facing message |
|---|---|
| `stt.enabled` is false | "voice transcription is not enabled on this gateway" |
| Configured model not downloaded | "STT model 'X' is not downloaded" |
| `sttTranscribe()` throws | "Couldn't transcribe your voice message. Please try again…" |
| `sttTranscribe()` returns empty string | "Couldn't make out anything in your voice message…" |

### Serialization

`sttChain` (a single Promise the next transcription chains off) prevents parallel `whisper-cli` runs from OOM-ing on memory-constrained hosts. Synchronous `this.sttChain = myTurn.catch(...)` reassignment chains arrivals in order with no race, and a failed transcription doesn't poison the next job.

A single "⏳ Transcribing a previous voice message — yours is queued" ack is sent when another transcription is already in flight, so the user isn't left silent for the duration × queue position. No count is surfaced — `sttPending` is checked as a boolean internally and decremented in the chain's `finally`.

## Verification

- `tsc --noEmit` clean on `packages/jimmy`
- Tested in production on a 1.9 GB Amazon Linux box with whisper.cpp v1.8.4 and ggml-small model — transcribes PT-BR and English voice notes, serializes concurrent voice messages without OOM, drops voice messages with the right user-facing message when STT is unavailable

### DIFF
--- a/packages/jimmy/src/connectors/telegram/index.ts
+++ b/packages/jimmy/src/connectors/telegram/index.ts
@@ -40,6 +40,7 @@ export class TelegramConnector implements Connector {
 
   private readonly sttConfig?: TelegramConnectorConfig["stt"];
   private sttChain: Promise<unknown> = Promise.resolve();
+  private sttPending = 0;
 
   constructor(config: TelegramConnectorConfig) {
     this.bot = new TelegramBot(config.botToken, { polling: false });
@@ -139,6 +140,20 @@ export class TelegramConnector implements Connector {
         const language = langs.length === 1 ? langs[0] : "auto";
 
         // Serialize transcriptions: parallel whisper-cli runs OOM on small hosts.
+        // If another transcription is already in flight, send a one-line ack so
+        // the user doesn't sit through ~duration × queue position in silence.
+        this.sttPending++;
+        if (this.sttPending > 1) {
+          try {
+            await this.bot.sendMessage(
+              telegramMsg.chat.id,
+              "⏳ Transcribing a previous voice message — yours is queued.",
+            );
+          } catch {
+            /* non-fatal */
+          }
+        }
+
         const myTurn = this.sttChain.then(async () => {
           try {
             await this.bot.sendChatAction(telegramMsg.chat.id, "typing");
@@ -163,7 +178,11 @@ export class TelegramConnector implements Connector {
             }
           }
         });
-        this.sttChain = myTurn.catch(() => undefined);
+        this.sttChain = myTurn
+          .catch(() => undefined)
+          .finally(() => {
+            this.sttPending = Math.max(0, this.sttPending - 1);
+          });
 
         let transcript: string | undefined;
         try {
--- a/packages/jimmy/src/connectors/telegram/index.ts
+++ b/packages/jimmy/src/connectors/telegram/index.ts
@@ -40,7 +40,6 @@ export class TelegramConnector implements Connector {
 
   private readonly sttConfig?: TelegramConnectorConfig["stt"];
   private sttChain: Promise<unknown> = Promise.resolve();
-  private sttQueueDepth = 0;
 
   constructor(config: TelegramConnectorConfig) {
     this.bot = new TelegramBot(config.botToken, { polling: false });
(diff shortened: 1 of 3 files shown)


hristo2612:
Nice — this uses the right primitive (`stt/transcribe`) and does not reinvent anything. Confirmed `transcribe()` has no internal concurrency guard and nothing else serializes whisper-cli, so the `sttChain` queue is genuinely justified for memory-constrained hosts. Approving in principle; two small things before merge:

1. **Empty-text fall-through.** When `stt.enabled` is false or `getModelPath()` returns null, a voice message falls through with `text: ""` — which is the exact "No deferred tool marker" crash from #54. Could we drop or annotate voice messages when STT cannot run, instead of forwarding empty text?
2. **Optional KISS trim.** The "⏳ Queued — N ahead" messaging is a fair bit of logic for marginal UX. I am happy to merge with just the silent serialization if you would rather keep it lean — your call.

(Heads up: this and #60 edit the same hunk of `telegram/index.ts`, so whichever lands second will need a rebase to combine the two `messageText` blocks.)

jeffgardnerdev:
Both addressed in 8ea5051:

1. Restructured the voice path so any case where STT can't run (`stt.enabled` false, model not downloaded, transcription empty, or transcription threw) now sends the user a Telegram explanation and `return`s — no more `text:""` falling through to the handler.

2. Dropped `sttQueueDepth` and the queued-status messaging. Silent serialization via `sttChain` is unchanged so parallel `whisper-cli` runs still can't happen; users see the typing indicator once their turn starts. Net -10 lines.

Acked on the #60 overlap — I'll rebase whichever lands second to combine the two `messageText` blocks.

hristo2612 (changes requested):
Thorough review done — thank you for this, the implementation itself is solid. `tsc --noEmit` is clean and the full suite (210 tests) passes on this branch. The promise-chain serialization is genuinely well done: the synchronous `this.sttChain = myTurn.catch(...)` reassignment means concurrent messages chain in arrival order with no race, a failed transcription doesn't poison the next job, and the caller still sees the real error. Tempdirs are cleaned in `finally` on both paths, and the `file_id`-keyed download has no SSRF surface. Nice work.

Two things before this can merge:

**1. (Blocking) The PR description doesn't match the code.** The body describes a `sttQueueDepth` counter and a user-facing "queued, N ahead" status message. Neither exists in the diff — they were removed in the "simplify serialization" commit. Please either rewrite the description to match what the code actually does, or restore the queue-depth feedback if you intended to keep it. As a side note: with no "N ahead" message, a user sending several voice notes will wait silently (up to ~duration × queue position) with zero feedback — worth a one-line status ack even without the counter.

**2. (Coordination with #60)** Both this PR and #60 handle `video_note`. If both land as-is, a video note gets *transcribed AND attached as a raw .mp4* — duplicate handling. Since these conflict in `telegram/index.ts` anyway, the plan is: we merge #60 first, then you rebase this PR, resolve the conflict (keep both import groups + both handler blocks, use `attachments` not `[]`), and drop `video_note` from #60's spec list here so it's transcribed only. Transcription is the more useful behavior for a voice-style clip.

**Nice to have:** a test mocking `downloadFile`/`sttTranscribe` (transcript→text, failure→drop, two concurrent→serialized order).

jeffgardnerdev:
Both addressed:

- **Description rewritten** to match the actual code — four explicit fall-through paths listed in a table, serialization section accurate, no more references to the removed `sttQueueDepth` counter.
- **Queue silence gap closed in 9de1f3b** — added a one-line "⏳ yours is queued" ack when another transcription is already in flight. `sttPending` is checked as a boolean (no "N ahead" count exposed), decremented in the chain's `.finally()`. Typecheck clean.

Ack on the #60 / `video_note` coordination plan — I'll handle the rebase once #60 lands.

PR #60, opened 2026-05-19
feat(telegram): expose file attachments via msg.attachments

### LINKED ISSUE #55: Telegram connector ignores document/photo/video attachments — files vanish silently
### Summary

The Telegram connector in `packages/jimmy/src/connectors/telegram/index.ts` declares `attachments: true` in its capabilities but only forwards `telegramMsg.text`. When a user attaches a document, photo, or video, the file is never downloaded, never persisted, never surfaced to the engine — it's silently dropped. There's no error, no warning, no user feedback. The user sends a PDF, the assistant reacts as if nothing arrived.

The repo also already has a fully wired `/api/files` endpoint and `~/.jinn/files/` storage with SQLite registration (`insertFile`/`listFiles` in `packages/jimmy/src/sessions/registry.ts`). All the file-handling plumbing exists — the connector just doesn't use it.

This is closely related to issue [#54](https://github.com/hristo2612/jinn/issues/54) (voice-message STT wiring). Different functional area, same root cause: the Telegram connector treats any non-text message body as if it didn't exist.

### Reproduction

1. Attach a PDF (or image, or video) to a Telegram message and send it to the bot.
2. The assistant receives an empty `text` and no attachment metadata. Behavior depends on what the engine does with empty messages — in my setup it intermittently triggers the same "No deferred tool marker" crash seen in [#54](https://github.com/hristo2612/jinn/issues/54) when the empty message reaches Claude session resume.
3. Nothing appears in `~/.jinn/files/` or in `GET /api/files`.

### Proposed fix

Detect `document`, `photo`, and `video` on incoming messages, download via the bot API, persist to the managed `FILES_DIR`, register via `insertFile()`, and append a marker line to the message text so the engine sees the attachment.

```ts
import crypto from "node:crypto";
import fs from "node:fs";
import os from "node:os";
import path from "node:path";
import { FILES_DIR } from "../../shared/paths.js";
import { insertFile } from "../../sessions/registry.js";

// inside bot.on("message", ...) handler, after voice/STT block, before building `msg`:

let messageText = telegramMsg.text || telegramMsg.caption || "";

const attachmentSpecs: Array<{
  file_id: string;
  suggestedName?: string;
  mime?: string;
  kind: "document" | "photo" | "video";
}> = [];

if (telegramMsg.document) {
  attachmentSpecs.push({
    file_id: telegramMsg.document.file_id,
    suggestedName: telegramMsg.document.file_name,
    mime: telegramMsg.document.mime_type,
    kind: "document",
  });
}
if (telegramMsg.photo && telegramMsg.photo.length > 0) {
  // largest size is last in Telegram's array
  const largest = telegramMsg.photo[telegramMsg.photo.length - 1];
  attachmentSpecs.push({
    file_id: largest.file_id,
    mime: "image/jpeg",
    kind: "photo",
  });
}
if (telegramMsg.video) {
  attachmentSpecs.push({
    file_id: telegramMsg.video.file_id,
    suggestedName: telegramMsg.video.file_name,
    mime: telegramMsg.video.mime_type || "video/mp4",
    kind: "video",
  });
}

const savedAttachments: Array<{
  id: string;
  displayName: string;
  path: string;
  kind: string;
  size: number;
}> = [];

for (const spec of attachmentSpecs) {
  try {
    fs.mkdirSync(FILES_DIR, { recursive: true });
    const tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "tg-file-"));
    try {
      const downloaded = await this.bot.downloadFile(spec.file_id, tmpDir);
      const id = crypto.randomUUID();
      const ext =
        (spec.suggestedName && path.extname(spec.suggestedName)) ||
        path.extname(downloaded) ||
        "";
      const finalPath = path.join(FILES_DIR, `${id}${ext}`);
      fs.copyFileSync(downloaded, finalPath);
      const stat = fs.statSync(finalPath);
      const displayName =
        spec.suggestedName || `${spec.kind}-${id.slice(0, 8)}${ext}`;
      try {
        insertFile({
          id,
          filename: displayName,
          size: stat.size,
          mimetype: spec.mime || "application/octet-stream",
          path: finalPath,
        });
      } catch (regErr) {
        logger.warn(
          `[telegram] insertFile failed (continuing): ${regErr instanceof Error ? regErr.message : regErr}`,
        );
      }
      savedAttachments.push({ id, displayName, path: finalPath, kind: spec.kind, size: stat.size });
      logger.info(
        `[telegram] Saved ${spec.kind} '${displayName}' (${stat.size} bytes) → ${finalPath}`,
      );
    } finally {
      try { fs.rmSync(tmpDir, { recursive: true, force: true }); } catch {}
    }
  } catch (err) {
    logger.error(
      `[telegram] Failed to save ${spec.kind}: ${err instanceof Error ? err.message : err}`,
    );
  }
}

if (savedAttachments.length > 0) {
  const block = savedAttachments
    .map((a) => `📎 ${a.kind}: ${a.displayName} (${a.size} bytes) → ${a.path}`)
    .join("\n");
  messageText = messageText ? `${messageText}\n\n${block}` : block;
}

// then use `messageText` when building IncomingMessage
```

The marker-line format gives the engine both a human-readable name and an absolute path it can read directly. Captions are preserved (since photos and videos commonly carry one).

### Notes / open questions

- Tested end-to-end against `0.10.0` dist. Sent a 729 KB PDF from Telegram → arrived in `~/.jinn/files/` with a valid UUID name, registered in SQLite, listed by `GET /api/files`, and the engine saw the marker line and could read the file.
- Telegram bot API caps inbound `getFile` downloads at **20 MB**. Larger files come back as a `file is too big` error from `bot.downloadFile()`. The fix above logs and continues; for really large files, users would need a separate path (portal upload). Worth surfacing this limit to the user as a friendly message rather than a silent failure.
- `IncomingMessage` already has an `attachments: []` field that's never populated. The marker-line approach is engine-agnostic and works with the current type, but a cleaner long-term fix would also push structured `{ id, filename, mimetype, path }` objects into `attachments[]` so consumers don't have to parse text. I left the array empty in the prototype to keep the change small.
- Same architectural question as [#54](https://github.com/hristo2612/jinn/issues/54): the WhatsApp connector should get parallel treatment (Baileys exposes attachments natively); Slack and Discord pass URLs differently and probably need their own approaches.
- Happy to send a PR, ideally bundled with the [#54](https://github.com/hristo2612/jinn/issues/54) fix since they share the message-handler restructuring.

### Environment

- `jinn-cli@0.10.0`
- Node 24.15.0
- Amazon Linux 2023

jeffgardnerdev:
Closes #55.

## Problem

The Telegram connector silently drops files — documents, photos, videos, video notes, animations, stickers all arrive at the bot but never reach the engine. From the user's perspective the file just vanishes.

## Change

Follow the existing Slack/WhatsApp attachment pattern:

- Detect attachment-bearing Telegram updates: `document`, `photo`, `video`, `video_note`, `animation`, `sticker`
- Download via `bot.downloadFile()` to `TMP_DIR`
- Rename to `<uuid><ext>` so repeated Telegram basenames don't collide (matches Slack's `downloadAttachment` helper)
- Push `{ name, url, mimeType, localPath }` onto `msg.attachments`

That's the whole change — `sessions/manager.ts:177` already extracts `localPath` and the engines auto-inject "Attached files: …" into the prompt, so no bespoke text formatting or `FILES_DIR` plumbing is needed.

## Verification

- `tsc --noEmit` clean on `packages/jimmy`
- Tested with PDFs, JPEG photos, MP4 videos, animated GIFs, video notes, and stickers — all reach the engine via `msg.attachments`

### DIFF
--- a/packages/jimmy/src/connectors/telegram/index.ts
+++ b/packages/jimmy/src/connectors/telegram/index.ts
@@ -1,4 +1,5 @@
 import TelegramBot from "node-telegram-bot-api";
+import { randomUUID } from "node:crypto";
 import fs from "node:fs";
 import path from "node:path";
 import type {
@@ -147,12 +148,19 @@ export class TelegramConnector implements Connector {
         fs.mkdirSync(TMP_DIR, { recursive: true });
         for (const spec of specs) {
           try {
-            const localPath: string = await (this.bot as any).downloadFile(
+            const downloaded: string = await (this.bot as any).downloadFile(
               spec.file_id,
               TMP_DIR,
             );
+            // Rename to a UUID so repeat Telegram basenames don't collide
+            // (matches the Slack connector's downloadAttachment pattern).
+            const ext =
+              path.extname(downloaded) ||
+              (spec.name ? path.extname(spec.name) : "");
+            const localPath = path.join(TMP_DIR, `${randomUUID()}${ext}`);
+            fs.renameSync(downloaded, localPath);
             attachments.push({
-              name: spec.name || path.basename(localPath),
+              name: spec.name || path.basename(downloaded),
               url: localPath,
               mimeType: spec.mime || "application/octet-stream",
               localPath,
--- a/packages/jimmy/src/connectors/telegram/index.ts
+++ b/packages/jimmy/src/connectors/telegram/index.ts
@@ -1,4 +1,8 @@
 import TelegramBot from "node-telegram-bot-api";
+import crypto from "node:crypto";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
 import type {
   Connector,
   ConnectorCapabilities,
@@ -11,6 +15,8 @@ import type {
 import { deriveSessionKey, buildReplyContext, isOldTelegramMessage } from "./threads.js";
 import { formatResponse } from "./format.js";
 import { logger } from "../../shared/logger.js";
+import { FILES_DIR } from "../../shared/paths.js";
+import { insertFile } from "../../sessions/registry.js";
 
 export class TelegramConnector implements Connector {
   name = "telegram";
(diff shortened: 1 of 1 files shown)


hristo2612:
The feature is wanted, but I would like to rework the approach before merging — right now it reimplements an attachment pipeline the codebase already has.

Slack, Discord, and WhatsApp all download a file locally and push `{ name, url, mimeType, localPath }` onto `msg.attachments` (see `slack/index.ts:148`, `whatsapp/index.ts:276`). From there `sessions/manager.ts:177` pulls out `localPath` and every engine auto-appends "Attached files: …" to the prompt (`claude.ts:87`, etc.). So the whole thing is: download → push to `msg.attachments`. This PR instead leaves `attachments: []` empty and rebuilds all of that by hand — copying into `FILES_DIR` with UUID names, writing rows into the web-upload `files` table (unrelated to chat attachments), and manually formatting a `📎` text block that duplicates the engine's own injection.

Suggested direction: follow the Slack/WhatsApp pattern — download each Telegram file to a temp/local path, push an `Attachment` onto `msg.attachments`, and delete the bespoke `FILES_DIR` + `insertFile` + text-injection code. Should drop this from ~117 lines to ~20 and behave consistently with the other connectors.

Also, the description and code diverge: it says it handles `video_note`, `animation`, and `sticker`, but only `document`/`photo`/`video` are in `attachmentSpecs` (so `video_note`/`sticker` still silently vanish — the original #55 bug); and the on-disk file is a UUID, not the "subdirectory keyed by `file_unique_id`" / preserved filename the description claims. Worth reconciling those too.

jeffgardnerdev:
Rewritten in 13a6e90 to follow the Slack/WhatsApp pipeline — `bot.downloadFile` → push onto `msg.attachments` → `sessions/manager.ts` extracts `localPath` and the engine auto-injects "Attached files: …". Dropped the `FILES_DIR` copy, `insertFile` write, UUID rename, and the manual 📎 text block.

Also added `video_note`, `animation`, and `sticker` to the spec list — those three were silently vanishing in the original branch, so the description-code divergence you flagged is now resolved (PR body updated to match).

Net -40 lines on the attachment block.

hristo2612 (commented):
Reviewed — this is clean, focused, and correctly mirrors the existing Slack attachment pattern. `tsc --noEmit` clean, 210 tests pass. Per-file try/catch so one bad download doesn't sink the batch. I'm happy to merge this first, then have #59 rebase on top.

Minor, non-blocking:

- **Title/body say `FILES_DIR`, but the code writes to `TMP_DIR`.** Just a stale description — please fix the wording (the code is right; `TMP_DIR` matches the Slack precedent).
- **Filename collisions:** you delegate naming to the library, which reuses Telegram's basenames (e.g. `documents/file_12.pdf`), so repeated files can overwrite each other in `TMP_DIR`. Slack avoids this with `randomUUID()`. Low impact since the engine reads `localPath` immediately, but it's a divergence from the Slack pattern worth matching.
- Heads-up: `video_note` is also handled by #59 (transcription). To avoid double-handling, we'll drop `video_note` from the spec list here during #59's rebase — no action needed from you on this PR.

Fix the `FILES_DIR`→`TMP_DIR` title and I'll merge.

jeffgardnerdev:
Both addressed:

- Title renamed and body refreshed to drop the stale `FILES_DIR` framing.
- 2559c50 renames downloaded files to `<uuid><ext>` before pushing to `msg.attachments`, matching Slack's `downloadAttachment` helper. Repeat Telegram basenames no longer collide in `TMP_DIR`. Typecheck still clean.

Ack on `video_note` — happy to leave that to your #59 rebase.

PR #78, opened 2026-07-18
fix: cold-resume retry for silently-unsubmitted warm-PTY paste

cheeselvl-star:
## Bug

Attaching a file/image to a message in an already-warm chat session (i.e. a follow-up turn, not the session's first turn) can hang the turn forever with no response and no error surfaced anywhere.

## Root cause

The warm-PTY prompt injection path uses Claude Code's interactive bracketed-paste-and-submit sequence. When the composed text carries a newline, that submit can silently fail — the text lands in the TUI's input box but the Enter never actually submits it. Every attachment-bearing prompt hits this, because the injected "Attached files:\n- <path>" suffix always embeds a newline. When it happens: no Stop hook fires, no SSE/transcript activity is produced, and the turn just hangs — indistinguishable from a legitimately slow response until the user gives up.

## Fix

- Arm a 20s watchdog immediately after a warm-PTY inject.
- If zero hook/SSE activity has been observed by the deadline, treat it as an unsubmitted paste: kill the stuck PTY and cold-resume the same turn exactly once via `spawn()` (prompt passed through argv instead of interactive paste — doesn't hit this failure mode).
- Release the dead PTY's lifecycle entry *synchronously* right after the kill, not just via the async `onExit` cleanup path. The recursive retry re-enters `run()` before that async cleanup would have completed, so without the synchronous release, `getWarm()` still returns the dying handle and the retry re-injects into a corpse instead of cold-spawning. `releaseSession()` is idempotent/identity-gated, so this doesn't race the PTY's own later `onExit` teardown.

## Testing

Reproduced and verified in an isolated environment (separate port/data dir, same vendor v0.25.0 base):
- 3 pre-fix captures of the broken behavior (watchdog fires, but the retry silently re-injects into the dead PTY and settles empty within ~400ms — no user-visible error, just a response-less "completed").
- 2 clean post-fix reproductions: warm chat → attach image → 20s watchdog → real cold `--resume` spawn logged within ~10ms of the kill → real assistant response describing the attached image rendered in the client.
- Sanity-checked native slash commands (e.g. `/usage`) are unaffected.

No behavior change for the non-buggy path (turns that submit normally never arm/trip the watchdog).

### DIFF
--- a/packages/jinn/src/engines/claude-interactive.ts
+++ b/packages/jinn/src/engines/claude-interactive.ts
@@ -423,6 +423,19 @@ const BACKGROUND_CLEAR_QUIET_MS = 10_000;
 
 const NATIVE_COMMAND_QUIET_MS = 1800;
 const NATIVE_COMMAND_MIN_MS = 3000;
+
+/** A warm-PTY prompt injection (bracketed-paste + delayed \r, see pasteAndSubmit)
+ *  can silently fail to submit — most reliably reproduced with multi-line prompts
+ *  (any prompt carrying attachments, since the "Attached files:\n- <path>" suffix
+ *  always embeds a newline). The Claude Code TUI renders an image-path paste as an
+ *  attachment chip, and the composed input is left sitting unsubmitted with no Stop
+ *  hook ever firing and no transcript activity — the turn hangs forever with no
+ *  error surfaced. If a warm-injected turn produces zero hook/SSE activity within
+ *  this window, treat it as an unsubmitted paste: kill the stuck PTY and cold-resume
+ *  once (spawn() passes the prompt via argv instead of interactive paste, which does
+ *  not hit this failure mode). */
+const STUCK_PASTE_TIMEOUT_MS = 20_000;
+const STUCK_PASTE_SENTINEL = "Interrupted: unsubmitted paste (no CLI activity within timeout)";
 const NATIVE_COMMAND_MAX_MS = 90_000;
 const LOST_STOP_RECOVERY_QUIET_MS = 60_000;
 const LOST_STOP_RECOVERY_MIN_MS = 5 * 60_000;
@@ -473,7 +486,7 @@ export class InteractiveClaudeEngine implements InterruptibleEngine, PtyViewEngi
    *  released by a kill->respawn race can't poison the freshly-started turn.
    *  `onStream` is the current turn's delta callback; the per-PTY SSE proxy routes
    *  parsed events here (a PTY outlives its turn, so the proxy looks this up live). */
-  private active = new Map<string, { resolver: TurnResolver; onStream?: (d: StreamDelta) => void; boundProc?: pty.IPty }>();
+  private active = new Map<string, { resolver: TurnResolver; onStream?: (d: StreamDelta) => void; boundProc?: pty.IPty; sawActivity?: boolean }>();
   /** Sessions with an in-flight async idle-spawn (proxy.start awaited) — prevents
    *  a second ensureIdleSpawn from racing in a duplicate PTY during that gap. */
   private idleSpawning = new Set<string>();
@@ -502,6 +515,9 @@ export class InteractiveClaudeEngine implements InterruptibleEngine, PtyViewEngi
   private backgroundActivityCb?: (jinnSessionId: string, info: UpstreamActivityInfo | null) => void;
   /** Test override for the post-settle clear quiet window (default 10s). */
   backgroundClearQuietMs = BACKGROUND_CLEAR_QUIET_MS;
+  /** Sessions currently retrying after a stuck-unsubmitted-paste kill — caps the
+   *  auto cold-resume to one attempt per originating turn (see STUCK_PASTE_SENTINEL). */
+  private stuckPasteRetried = new Set<string>();
 
   constructor(
     private lifecycle: PtyLifecycleManager,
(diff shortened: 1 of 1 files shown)


hristo2612:
Thanks for digging into this — the diagnosis and the cold-resume recovery path look plausible, and the synchronous lifecycle release is an important detail.

I auto-merged the PR onto current `main`/v0.26 locally: there are no textual conflicts, typecheck/build pass, and the full Jinn package suite passes (3,558 passed, 1 skipped). I have **not** personally reproduced the attachment flow end-to-end on v0.26 yet.

Before merging, could you rebase onto current `main` and tighten this toward a KISS implementation?

The main correctness concern is that “no hook/SSE event for 20s” does not prove the paste was never submitted. An upstream request may already be active but have a slow first SSE event; v0.26 tracks that via `hasActiveUpstream()`, but the watchdog does not consult it. Killing and cold-resuming in that state could submit the same user turn twice.

My preferred smallest fix would be: for a warm turn containing attachments (and multiline prompts too, if your reproduction shows they are all affected), release the warm PTY and cold-resume immediately through the existing spawn path. That avoids the unreliable paste path without a 20s delay, sentinel error, retry set, or recursive `run()`.

If preserving warm multiline submission matters, Claude Code’s `UserPromptSubmit` hook would be a much stronger acknowledgement than waiting for SSE output. At minimum, the watchdog should never fire while upstream activity is present.

Could you also add an automated regression test on rebased v0.26 covering:

- stuck warm attachment paste → exactly one cold resume with the original prompt/attachment;
- submitted request with slow/no initial SSE but active upstream → no retry/duplicate;
- stale killed-PTY exit → does not poison the replacement PTY.

Small cleanup: `entry.boundProc?.kill()` followed by `releaseSession()` kills twice; `releaseSession()` already handles the kill.

Your end-to-end reproduction notes are encouraging. If you can rerun that flow after rebasing and include the regression test, I think this will be in much stronger shape.

hristo2612:
Just checking in on this one — no rush at all. Since you had a solid end-to-end reproduction, were you able to narrow down *why* the warm paste silently doesn't submit (e.g. does Claude Code's `UserPromptSubmit` hook ever fire in the stuck case, or is the `\r` genuinely being dropped)? Even a rough finding would help us decide between the watchdog-recovery approach and just cold-resuming attachment/multiline prompts up front. Really appreciate the digging you've already done here 🙏

hristo2612:
Thank you for this — the bug is real, your reproduction was what made it findable, and it's **fixed in v0.28.6**. But the root cause turned out to be different from the one in the description, so the fix landed as a one-line change rather than this patch. Full evidence below, since you did the legwork and deserve to see why.

**Newlines aren't the trigger.** I replayed jinn's exact injection bytes into a real `claude` PTY across a 13-run matrix. Two rows falsify the newline hypothesis outright:

| newlines | image path | submits? |
|---|---|---|
| 3 | `/tmp/image.png` (**doesn't exist**) | ✅ |
| **0** | real 5.8 MB PNG | ❌ **hangs** |
| 3 | real 8 MB **.txt** | ✅ |
| 3 | real 5.8 MB PNG | ❌ hangs |
| 3 | same PNG, CR at **+400 ms** | ✅ |
| 3 | same PNG, path in **backticks** | ✅ |

**The real mechanism:** Claude Code's TUI scans bracketed-pasted text for tokens that resolve to an existing **image** file. On a hit it enters an async "Pasting…" state while it reads and base64-encodes the file into an `[Image #N]` chip, and **discards keypresses while that state is active** — including the submit CR. `pasteAndSubmit` fires that CR on a fixed 150 ms timer, and any real screenshot takes longer than 150 ms to encode, so the CR is swallowed.

Timeline from the failing capture (ms after the paste write; CR was written at +150):

```
+162  "…Pasting…"                     <- TUI enters attach state
+225  "…[Image #1]Reply with…"        <- renders the chip
+242  "…⏸ manual mode on…"
+7395 (silence; never submits)
```

Your newline observation was a passenger: every attachment prompt carries both a newline *and* an image path, and the image path is the driver. That also explains the "intermittent" feel — the window scales with image size and machine load.

Notably, flushing `\x1b[201~\r` in the **same write** does *not* help (the CR is consumed inside the same handler tick), so the fix genuinely had to remove the async state rather than re-time the CR.

**The fix** is to stop handing the TUI a bare, auto-attachable path — attachment paths are now backticked, so nothing auto-attaches and there's no state to race. Verified submitting at the unchanged 150 ms. It also converges warm behaviour onto the cold/argv path, which never auto-attached, so the model resolves the path with `Read` (which renders images) in both cases. Locked by a unit test asserting the composed bytes contain no bare image path — no live PTY needed.

**Why not this patch:** the 20 s watchdog + kill + cold-resume treats the symptom, adds 20 s of latency to every affected turn, and leaves the paste still racing. Your synchronous `releaseSession()` observation is **accurate** — `onExit`'s identity-gated release is async, so `getWarm()` can briefly return a dying handle — but it's only load-bearing for the kill/retry design, and nothing in the new fix kills a PTY.

One thing your report surfaced that I'd have missed: this isn't attachment-only. `writeStdin` shares `pasteAndSubmit`, so a human pasting an image path into the web terminal view hit the identical hang. That's fixed too.

Could you re-verify on v0.28.6 and close this if it holds for you? And thanks again — the isolated-environment captures were genuinely useful.

hristo2612:
**Correction to my previous comment.** I wrote:

> One thing your report surfaced that I'd have missed: this isn't attachment-only. `writeStdin` shares `pasteAndSubmit`, so a human pasting an image path into the web terminal view hit the identical hang. That's fixed too.

The last sentence is **wrong**, and I'd rather say so than have you re-verify against a false claim.

The backticking lives in `buildAttachmentSuffix()`, which is only reached from the two prompt-composition paths — `buildInteractiveArgs` (cold/argv) and `injectPrompt` (warm PTY). `writeStdin` doesn't call it:

```ts
writeStdin(sessionId: string, text: string): void {
  const handle = this.lifecycle.getWarm(sessionId);
  if (!handle) return;
  const proc = (handle as any)._proc as pty.IPty | undefined;
  if (!proc) return;
  pasteAndSubmit(proc, text);   // raw, un-backticked
}
```

So **pasting a real image path into the web CLI/terminal view still reproduces the original hang on v0.28.6.** The observation you surfaced was right; my claim that it was covered was not. The identification of the shared `pasteAndSubmit` seam still stands entirely to your credit — I just didn't follow it through to the second caller.

Two related gaps found since, both still live on v0.28.6, so the mechanism you found has a wider blast radius than the fix currently covers:

1. A bare image path in the **body** of a normal chat message (not the attachment list) hits the same auto-attach race — `buildAttachmentSuffix` only backticks the "Attached files:" block, and nothing sanitises the body.
2. `writeStdin`, above.

Both are the same root cause you nailed. I'll fix them properly rather than patch the one path — tracking separately so this PR isn't held open for it. Sorry for the incorrect all-clear.

hristo2612:
Closing this, with thanks, because the bug it reports is fixed and the mechanism you asked about has since been built.

**The bug is fixed as of v0.28.6.** As I wrote above, the root cause turned out not to be newlines. My 13-run byte-replay matrix falsified that directly: a **zero-newline** prompt with a real 5.8 MB PNG hangs, while three newlines with a non-existent image path submits fine. It was Claude Code's UI auto-attaching a pasted image path and swallowing keystrokes while it encoded the file, including the Enter that submits. The fix was to pass attachment paths in a form the UI does not try to auto-attach, which removes the race rather than waiting on it.

**The stronger fix I suggested to you in July now exists.** My first review said the watchdog should never fire on `no SSE for 20s` alone, because that does not prove the paste was never submitted, and that Claude Code's `UserPromptSubmit` hook would be a much better acknowledgement than waiting for output. That is exactly what shipped in #94, released in **v0.29.0**: the submit is retried until the hook acknowledges it, gated on `hasActiveUpstream()` so an in-flight request cannot be double-submitted, with the stalled state surfaced in the sidebar instead of guessed at.

So both halves of what this PR was reaching for are now on `main`, by a different route, and this branch no longer merges cleanly.

None of that takes anything away from the report. **Your reproduction is what made this findable at all** — it was a silent hang with no error and no log line, and without a reliable way to trigger it there was nothing to investigate. The reason the eventual fix was one line is that you narrowed where to look.

Thank you for the digging, and please do open things like this again. 🙏

PR #79, opened 2026-07-19
fix(web): messages disappear from chat after turns in long sessions (stale turn-start index)

juyoungk23:
## Symptom

In sessions longer than the paged snapshot window (150 messages), previously rendered messages — both user and assistant, including media — disappear from the chat pane when a turn completes. The newest message often reappears on the next snapshot; older rows stay hidden until scroll-up paging. Nothing is lost in the DB; it is purely a client render bug.

## Root cause

`use-live-session.ts` tracks the current turn's start via `intermediateStartRef`. A reload effect fires `loadSession` shortly into every turn, and its running branch set:

```ts
intermediateStartRef.current = firstPartialIndex >= 0 ? firstPartialIndex : backendMessages.length
```

`firstPartialIndex`/`backendMessages.length` are indices into the **newest-150 snapshot**, but the ref is later applied to the **full rendered list** (which `mergePagedSnapshot` deliberately grows past 150 by keeping older rows). On `session:completed`, the handler does `list.slice(0, turnStart)` + re-appends media/tail — with the stale snapshot-space index this deletes the newest ~N rendered messages every turn completion (media rows survive the slice but get re-ordered to the bottom, which is why images appear to move).

Repro conditions: any session with >150 messages + streaming turns. Engine restarts amplify it (more rows slide the window faster) but are not the trigger.

## Fix

Anchor the turn start by **message id** instead of snapshot index: capture `backendMessages[firstPartialIndex].id` and remap it inside the `setMessages` updater via `findIndex` against whichever array is actually returned (merged or kept-current), falling back to `merged.length`. One file, +12/−2.

Verified on a 350-message live session: turn completions no longer truncate the rendered list; media rows stay in place. `tsc --noEmit` clean on the touched file (the 5 pre-existing repo errors are unrelated missing `@testing-library/user-event` types in tests).

### DIFF
--- a/packages/web/src/hooks/use-live-session.ts
+++ b/packages/web/src/hooks/use-live-session.ts
@@ -1054,15 +1054,25 @@ export function useLiveSession(
         // The server now persists mid-turn partial blocks, so the backend snapshot
         // already carries in-progress output — no localStorage replay needed. This is
         // what makes a mid-turn refresh restore the streamed blocks on any device.
-        intermediateStartRef.current = firstPartialIndex >= 0 ? firstPartialIndex : backendMessages.length
+        // The turn-start marker must be an index into the array actually rendered,
+        // not the (possibly ≤150-row) server snapshot — anchor it by message id and
+        // remap it inside the updater against whichever array we return. Otherwise a
+        // snapshot-space index truncates already-rendered history at turn completion.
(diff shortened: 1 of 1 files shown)


hristo2612:
Merged 🎉 Really clean diagnosis — pinning the turn-start to a stable message id instead of a snapshot-space index is exactly the right call, and keeping it to one file made this trivial to verify. Thank you for tracking down a genuinely subtle long-session render bug. Please keep them coming if you spot anything else!

PR #80, opened 2026-07-19
feat(limits): show all Claude usage buckets incl. model-scoped weekly limits (e.g. Fable)

juyoungk23:
## Problem

The Limits tab only ever shows the 5-hour session window and the all-models weekly window for Claude. Anthropic subscriptions now carry additional **model-scoped weekly buckets** (e.g. a separate weekly limit for Fable/Mythos-class models) that never appear — users can exhaust a model budget with no visibility.

Root cause is two layers below the UI (which already renders `engine.windows[]` generically):
1. `shared/engine-limits.ts` read only `rate_limits.five_hour`/`seven_day` from Claude Code statusline snapshots.
2. Claude Code's statusline payload itself only ever contains those two buckets — model-scoped limits structurally cannot arrive via that path.

## Change

`collectClaudeLimits` now queries the OAuth usage endpoint directly (token from the same local credential store Claude Code uses; nothing logged or persisted) and maps the `limits[]` array **generically** via a new `windowsFromClaudeUsage()`: `session` → "5h", `weekly_all` → "7d", `weekly_scoped` → "7d <model display_name>", unknown future kinds pass through by name; a fallback maps any named bucket with numeric `utilization`. The statusline snapshot remains the fallback source (offline/no token) and still supplies context-window/cost data. Env kill-switch: `JINN_CLAUDE_USAGE_API=off`.

## Verification

- 4 new unit tests for the mapping (scoped bucket, unknown kinds, fallback, malformed entries); existing tests pinned to the snapshot path via the kill-switch; all pass, plus full build.
- Live on a real account: tab shows `5h`, `7d`, and `7d Fable` with correct utilization and reset times — surfacing a model-scoped budget at 50% that was previously invisible.

### DIFF
--- a/packages/jinn/src/shared/__tests__/engine-limits.test.ts
+++ b/packages/jinn/src/shared/__tests__/engine-limits.test.ts
@@ -68,6 +68,9 @@ beforeAll(async () => {
   fs.mkdirSync(CLAUDE_DIR, { recursive: true });
   process.env.JINN_HOME = JINN_HOME_TMP; // frozen into paths.ts at first import below
   process.env.CODEX_HOME = CODEX_HOME_TMP; // read per-call by the collector
+  // Keep the collector off the live OAuth usage API (keychain + network) so
+  // these tests stay deterministic and exercise the statusline-snapshot path.
+  process.env.JINN_CLAUDE_USAGE_API = "off";
   ({ collectEngineLimits } = await import("../engine-limits.js"));
   ({ invalidateModelRegistry } = await import("../models.js"));
 });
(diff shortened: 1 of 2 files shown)


hristo2612:
Merged 🎉 The generic `limits[]` mapping is the part I appreciated most — new server-side buckets will surface with no further code changes, and the statusline fallback + `JINN_CLAUDE_USAGE_API=off` kill-switch keep it safe when the token or endpoint isn't available. Nice fill for a real visibility gap. Thanks again, and keep it up — happy to look at whatever you find next!

PR #90, opened 2026-07-25
fix(todos): make review rejections reachable and declared blocks durable

juyoungk23:
A reviewer has no way to say "rejected" — the only thing they can write is a status, and no status counts as a rejection, so nothing tracks how many times a piece of work has come back and a review can bounce between reviewer and producer indefinitely without ever reaching the operator. This adds a proper verdict surface (pass / fail / blocked) wired into the round-counting path that already exists, so a stuck review escalates to the operator at the cap instead of looping — and it stops the reconciler from erasing a blocker before anyone has seen it.

---

Two things in the Todos state machine that are built but unreachable.

**The review bounce can't be expressed.** `transition()` supports `in_review → executing` with `rounds++`, escalating to the operator at `effectiveMaxRounds`. But the only agent route, `POST /work-items/:id/status`, passes `manual: true`, and `transitions.ts` refuses a manual move into `executing` from anything but `backlog`/`assigned`. So that edge never fires: `rounds` stays `0` for an item's whole life, the round cap never applies, and reviewers reject by writing `in_review → blocked` instead — an uncounted edge, so a review loop never terminates.

Adds `POST /work-items/:id/review` and a `review_verdict` tool. Three verdicts, all routed to existing `transition()` options:

- `pass` → `done`
- `fail` → `executing` `{bounce}`
- `blocked` → `blocked` `{declared}`

`fail` requires the full findings list, so one pass costs one round instead of N. Agent `in_review → blocked` is refused with a pointer to the new route; producers still block from `executing`. The reviewer check is extracted so verdict authority and done-closure authority stay in sync.

**Declared blocks get erased.** `blocked` means two things: an agent saying "this needs a decision", and the reconciler saying "the last attempt failed". Derivation won, so a declared block vanished as soon as any linked session was in flight — a watchdog, a follow-up probe — and never reached the operator's queue.

`deriveWorkItemStatus` now honours a `blockDeclared` flag, the same guard already applied to `in_review`. Provenance comes from the newest block event via one indexed row. Existing blocks carry no marker and fall back to actor provenance, so old data behaves exactly as before.

No new statuses. Manifest budget `4911 → 5125` for the new tool.

### DIFF
--- a/packages/jinn/src/gateway/api.ts
+++ b/packages/jinn/src/gateway/api.ts
@@ -167,6 +167,7 @@ import {
 import {
   appendWorkItemEvent,
   createWorkItem,
+  effectiveMaxRounds,
   getWorkItem,
   getWorkItemBySourceRef,
   getWorkItemSpend,
@@ -1002,6 +1003,11 @@ function cronJobSummary(job: Record<string, unknown>, lastRun: unknown): Record<
 const WORK_ITEM_STATUSES: readonly WorkItemStatus[] = ['backlog', 'assigned', 'executing', 'in_review', 'done', 'blocked', 'escalated', 'cancelled'];
 const WORK_ITEM_SOURCES: readonly WorkItemSource[] = ['human', 'delegation', 'cron', 'workflow', 'session', 'connector', 'goal'];
 const AGENT_WORK_ITEM_TARGETS: readonly WorkItemStatus[] = ['executing', 'in_review', 'blocked', 'escalated', 'done'];
+
+/** The three outcomes of a review pass. `fail` is the counted round; it is the
+ *  verdict the /status route structurally could not express. */
+const REVIEW_VERDICTS = ['pass', 'fail', 'blocked'] as const;
+type ReviewVerdict = (typeof REVIEW_VERDICTS)[number];
 const VERIFY_MODES = ['trust', 'verify', 'thorough'] as const;
 const VERIFY_POLICY_KEYS = new Set(['mode', 'verifier', 'maxRounds']);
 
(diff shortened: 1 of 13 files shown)


hristo2612:
Reviewed this against current `main` (rebases clean, typecheck clean, tests pass modulo pre-existing environment failures on a baseline worktree). **My recommendation is to split it** — one half is a proven fix I'd take as-is, the other needs work.

## Take now: the declared-block half

This is real and I reproduced it on `main`. An agent that declares `blocked` while its own session is still `running` — the normal case, since the agent declares mid-turn — is flipped straight back to `executing` by the next reconcile sweep (≤20s):

```
MAIN AFTER RECONCILE: executing     # expected: blocked
```

`deriveWorkItemStatus` guards `in_review` but not `blocked`, so derivation wins and a declared blocker vanishes before it ever reaches the operator's queue. Sourcing `blockDeclared` from the newest block event, with actor-provenance fallback for pre-existing rows, is correct, cheap (one indexed row, only when blocked), backward-compatible, and well tested. `reconcile.ts` + `store.ts` + those tests would merge on their own.

## Needs work: the review-verdict half

**The premise is overstated.** "`rounds` stays 0 for an item's whole life" isn't accurate — `approvals.ts` already reaches the bounce edge (`reject` + `in_review` → `transition(executing)` with `rounds++`, escalating at the cap), and the Todos v2 UI is already wired to it as **"Send back"**. The real gap is narrower and worth stating correctly: approvals are never auto-attached, so an `in_review` item **with no pending approval** has no rejection verb available to an agent reviewer.

**The `fail` bounce doesn't survive the reconciler.** I verified this in a worktree: after `verdict: fail` → `executing` (rounds=1), the next `reconcileWorkItem` derives it straight back to `in_review`, because the producer's newest attempt receipt is still `succeeded`:

```
AFTER RECONCILE: in_review  changed=true  rounds=1
```

So the headline behaviour is half-true — `rounds` increments and escalation eventually fires, but the producer is never re-engaged and the item reappears in the review queue within one sweep. The PR fixes derivation-wins for `blocked` and leaves the identical hole open on the edge it exists to enable.

**Other blockers:**

- **No route-level tests.** `work-items-route.test.ts` is untouched: the new `POST /work-items/:id/review`, the verdict-authority check, and the new **403 on `/status` for `in_review → blocked`** are all untested. `transitions-bounce.test.ts` characterizes pre-existing `transition()` behaviour, not the new code.
- **Two review vocabularies.** `pass/fail/blocked` sits alongside the shipped `approve/reject` with a *different* authority model, and the web UI knows nothing about `/review`. Merging mid-Todos-v2 forks the review contract.
- **Manifest budget 4911 → 5125** in a file that enforces near-zero headroom, for a tool no UI consumes yet.
- Pre-existing but the new route adds a second door to it: `verdict: pass` → `done` leaves a `pending` approval row dangling, and `needsAttentionFor` has no status filter, so the closed Todo stays in the Attention inbox.

## Suggested path

1. Open a PR with just the `blockDeclared` half — I'll merge that.
2. For the verdict half: correct the premise in the description, make the `fail` bounce durable against the reconciler (or document the gap), add route-level tests, and let's settle whether Todos v2 should get a second vocabulary or whether `review_verdict` should front the existing approval-decide engine.

Incidentally your framing was independently vindicated today: I hit exactly this gap trying to close a Todo I didn't own — there's no agent-reachable route, so it had to hop back to the creating session. The problem is real; it's the `fail` edge and the vocabulary fork I want resolved first.

hristo2612:
Following up on the one thing I left vague — "make the `fail` bounce durable against the reconciler" isn't a fair ask without pointing at the mechanism, so here it is.

`deriveWorkItemStatus` treats `in_review` as sticky, but reaches it from the producer's newest attempt receipt:

```ts
if (current === 'in_review') return current;   // sticky once there
...
if (newest === 'succeeded') return 'in_review';
```

A `fail` verdict moves the item `in_review → executing`, which drops that stickiness. The producer's newest attempt receipt is still `succeeded` — the bounce didn't invalidate it — so the very next sweep hits the second line and derives `in_review` again. That's the whole loop, and it's why `rounds` increments while the producer is never re-engaged.

**The fix is the pattern you already wrote in this PR.** Two options:

1. **Mirror `blockDeclared`.** Source a bounce marker from the newest `to_status='executing'` event and guard derivation with it, exactly as you did for `blocked`. Lowest risk, consistent with the half I want to merge, same one-indexed-row cost.

2. **Supersede the receipt on bounce.** Semantically the cleaner statement — "this attempt was reviewed and not accepted", so it should stop counting as the newest `succeeded` — but it touches attempt semantics shared with the workflow paths, so it needs more care.

I'd take (1). It keeps both halves of this PR expressing the same idea: a declared transition outranks a derived one.

Worth noting this makes the two halves less separable than my earlier comment implied — the `blockDeclared` guard and a bounce guard are the same mechanism applied twice. If you'd rather keep them together in one PR built around "declared transitions win over derivation", with the verdict route tested, that's a coherent PR and I'd review it as one. The split was only to get the proven half in quickly; it isn't a requirement.

hristo2612:
Closing this in favour of `4ef3661a`, which takes the half of it I said I'd merge, plus the bounce fix from my follow-up comment. Both bugs, one mechanism, and the credit is in the commit body.

## What landed

`fix(todos): preserve declared transitions during reconciliation` — [`4ef3661a`](https://github.com/hristo2612/jinn/commit/4ef3661a162fd70882dbebc7bebc1121050f3890), three files, +127/−2.

**Your `blockDeclared` insight is the whole of it.** `deriveWorkItemStatus` now takes declaration provenance and refuses to overwrite a status a caller explicitly declared: `blocked` while the producing session is still `running`, and `executing` after a review bounce whose newest attempt receipt is still `succeeded`. Same one-indexed-row lookup you proposed, with the actor-provenance fallback for pre-existing rows.

Two refinements on top of your version, both from the follow-up:

- **The lookup is lazy.** It only runs when receipt derivation would actually overwrite the current status, so an ordinary reconcile does no extra query at all.
- **`isReviewBounceDeclared` reads the *newest* transition into `executing`.** An older bounce must not outlive a later start or unblock, or the item would be pinned to `executing` forever.

Verified independently rather than on the implementer's word: with the two guard lines removed, exactly the two new tests fail and the other 29 pass, so they bite for the stated reason and nothing else. Full suite 289 files / 3579 passing, typecheck clean.

## What I deliberately did not take, and why

The `POST /work-items/:id/review` route and the `pass/fail/blocked` vocabulary. Not because they're wrong — the gap you identified is real and I confirmed it from the other side, hitting it myself trying to close a Todo I didn't own. But adding a second review vocabulary with a different authority model alongside the shipped `approve/reject`, mid-Todos-v2, forks the contract, and the web UI knows nothing about `/review`. That's a design decision rather than a fix, and it deserves its own issue where the question is "should `review_verdict` front the existing approval engine, or replace it" rather than riding along with a reconciler bug.

I also left attempt receipts alone. Superseding the receipt on bounce is arguably the cleaner statement semantically, as I said before, but it touches semantics shared with the Workflow paths and wasn't needed to close the hole.

## Thank you

Three of the four issues you filed this week were real and precisely diagnosed, and this one came with a working mechanism attached, which is why the fix is a small commit rather than an investigation. The `blockDeclared` idea — that a declared transition is governance state and outranks anything derived from session transport — is now the rule at both sites, and it will be the rule at the third when the review route gets built.

Please do keep filing. If you want to take the review-vocabulary question as its own PR, say so and I'll open the issue and we can settle the design there first, so you're not writing code against an unmade decision again.

PR #94, opened 2026-07-26
fix(claude): confirm the prompt was submitted, and surface a stalled turn

e1010101:
A warm-PTY turn is submitted by bracketed-pasting the prompt and writing a CR 150ms later, and nothing checks that the CR landed. When it is swallowed, the text sits in the composer, the turn never starts, and the session is pinned at `running` with its queue frozen behind it — rendered in the dashboard exactly like healthy work. This confirms the submit instead of assuming it, and makes a stalled turn visible.

> Rebased onto current `main`. `88ddd669` landed the attachment-submit fix and a terminal stall backstop while this was open, so everything that overlapped is gone: no reconciler changes, and **nothing here settles a turn**.

---

**Backticking attachment paths fixed one cause; this covers the rest.** That fix removes the characterised trigger — the TUI's async image encode discarding keypresses — at the source, which is the right fix for it. But it is not the only way to lose the CR. On a live instance, two of the three wedges I traced had no attachment involved at all: a large paste into a TUI mid-redraw, and one sitting at `3% until auto-compact`. Each produced the same signature — a finished previous turn, `[Pasted text #N +M lines]` still in the composer, no spinner, no error, no Stop hook.

**The confirmation retries; it does not decide.** `UserPromptSubmit` is the CLI's acknowledgement, and the CR is re-sent until it arrives — a CR re-sent a second later lands where the 150ms one did not, which is where the value is. If the acknowledgement never comes, `onUnconfirmed` **logs a warning and nothing else**. It does not settle the turn:

- Ruling a turn dead from the absence of a signal is wrong when the signal is genuinely absent for live work: Claude Code queues a prompt behind a running turn and emits no `UserPromptSubmit` until it dequeues, so queued and swallowed are indistinguishable by acknowledgement alone. The retry loop pauses on `hasActiveUpstream() || activeTools > 0` rather than acting on that ambiguity.
- `shouldSettleStalledTurn` already owns the verdict, with far more evidence. Losing a CR costs that backstop's window rather than seconds.
- Because no error string is produced here, the `startsWith("Interrupted")` quiet-preempt path stays unreachable, and the composer-merge exposure stays at the backstop's existing rate rather than being made ~150× more likely.

The retry loop is cancelled when the turn settles for any reason — otherwise it would keep writing CRs into a PTY that has moved on to a different turn, submitting whatever that composer happens to hold.

Two paths are deliberately exempt. Cold spawn carries its prompt in argv, so it is submitted by construction. Native commands (`/compact`, `/clear`, `/model`) settle through `nativeCommandTimer` and need not emit the hook, so confirming them would fire spurious CRs at a prompt that already did its work — the same exclusion the lost-Stop recovery makes.

**`turnProgress()` reports, it does not decide.** It exposes last-progress time, whether the prompt was acknowledged, in-flight local tool count, and upstream activity. It exists so the state can be read from outside the engine.

**Because until now it could not be seen.** A wedged session and a hard-thinking one were pixel-identical, indefinitely: same pulsing blue dot, no elapsed time, no queue depth anywhere in the UI. Three separate wedges on one instance went unnoticed for hours each — the first for 3h18m with twelve inter-employee callbacks queued behind it — and were found only by querying the session database directly.

A stalled session's dot is now amber and *still*, since a pulse reads as working, labelled `no output for 51m` or naming the unaccepted-prompt case, which has a different remedy. Rows carry queue depth, so a backlog piling up behind a wedged employee is legible without opening anything.

**The staleness verdict is the client's, deliberately.** The gateway serializes `turnProgress` — `lastProgressAt` plus `awaitingSubmit` — for any live turn and passes no judgement. A wedged session emits no events, so nothing invalidates the sessions query; a server-computed verdict would only arrive if something unrelated triggered a refetch, and the feature exists to surface a session that generates no signals. The client holds the instant from turn start, while the session is still healthy, and `TURN_STALL_VISIBLE_MS` is applied against a local clock driven by one shared 15s `useSyncExternalStore` tick — rows with nothing in flight never subscribe, so a quiet list arms no timer. The row therefore turns amber with no network round-trip.

What stays server-side is `turnProgress: null` while a tool or upstream request explains the quiet. That is state rather than time, and both edges of it emit hooks — hence events, hence a refetch — so the clearing signal is guaranteed.

**Tests.** At the `pasteAndSubmit` level: the retry loop, the cancellation contract, the settled-turn guard, the busy pause, and report-not-settle. Through `engine.run`: that a real `UserPromptSubmit` on the hook callback sets `promptSubmitted` while `SessionStart` does not, the cold-spawn and native-command exemptions, `turnProgress()` from the real engine, and that a turn settled unacknowledged writes no further CRs — each mutation-checked so none passes vacuously. Server-side: `turnProgress` reported 5s into a turn, null while a tool or upstream request is in flight, null for non-running sessions and engines that cannot report. Client-side: the threshold applied locally, one payload from turn start crossing it unaided and still advancing an hour later, and the shared clock arming no timer when idle.

One guarantee is deliberately left untested and marked as such in the test file: that `run()`'s `finally` calls `cancelSubmitConfirm()`. `submitted()` also reads `resolver.isSettled`, so removing that call changes no observable behaviour — verified by mutation — and it remains as redundancy rather than as an untested claim.

### DIFF
--- a/packages/jinn/src/engines/__tests__/claude-interactive-submit-wiring.test.ts
+++ b/packages/jinn/src/engines/__tests__/claude-interactive-submit-wiring.test.ts
@@ -1,0 +1,219 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+
+/**
+ * The submit confirmation, exercised through `engine.run` rather than by calling
+ * `pasteAndSubmit` with a hand-stubbed `submitted()`.
+ *
+ * The mechanism is covered at the `pasteAndSubmit` level in claude-interactive.test.ts;
+ * what is only reachable here is the wiring: that a real hook arriving on the
+ * registry callback flips `promptSubmitted`, that `SessionStart` deliberately does
+ * not, that the cold-spawn and native-command paths are exempt, and that a
+ * settled turn writes no further CRs. Getting those wrong makes the confirmation
+ * either useless (a false ack strands the prompt in the composer, the failure
+ * this exists to catch) or actively harmful (CRs written into a PTY that has
+ * moved on to a different turn).
+ *
+ * Every assertion here was mutation-checked against the production code, so none
+ * of them passes vacuously. See the note at the end for the one guarantee that
+ * cannot be pinned this way.
+ */
+
+interface FakePty {
+  pid: number;
+  _exitCode: number | null;
+  _exitCb?: (e: { exitCode: number }) => void;
+  writes: string[];
+  onData: (cb: (d: string) => void) => void;
+  onExit: (cb: (e: { exitCode: number }) => void) => void;
+  kill: (signal?: string) => void;
+  write: (d: string) => void;
+  resize: (c: number, r: number) => void;
+  on: (event: string, cb: (...a: any[]) => void) => void;
+  fireExit: () => void;
+}
+
+const ptys: FakePty[] = [];
+function makeFakePty(): FakePty {
+  const p: FakePty = {
+    pid: 2000 + ptys.length,
+    _exitCode: null,
+    writes: [], // records instead of discarding — the CRs are the subject here
+    onData() {},
+    onExit(cb) { p._exitCb = cb; },
+    kill() {},
+    write(d: string) { p.writes.push(d); },
+    resize() {},
+    on() {},
+    fireExit() { p._exitCode = 0; p._exitCb?.({ exitCode: 0 }); },
+  };
+  return p;
+}
+
+vi.mock("node-pty", () => ({
+  spawn: vi.fn(() => { const p = makeFakePty(); ptys.push(p); return p; }),
+}));
+vi.mock("../sse-pty-proxy.js", () => ({
+  MAIN_AGENT_SENTINEL: "<!-- jinn-main-agent:5c1f -->",
+  SsePtyProxy: class {
+    port = 0;
+    constructor(_label: string, _onEvent: (e: unknown) => void) {}
+    async start() { return 41100; }
+    stop() {}
+  },
+}));
+vi.mock("../shared/claude-settings.js", () => ({
+  writeSessionSettings: () => "/tmp/fake-settings.json",
+}));
+
+import { InteractiveClaudeEngine } from "../claude-interactive.js";
+import { PtyLifecycleManager } from "../pty-lifecycle.js";
+
+const CR = "\r";
+const crs = (p: FakePty) => p.writes.filter((w) => w === CR).length;
+
+describe("InteractiveClaudeEngine — submit confirmation wiring", () => {
+  let lifecycle: PtyLifecycleManager;
+  let hookCb: ((h: any) => void) | undefined;
+  let engine: InteractiveClaudeEngine;
+
+  beforeEach(() => {
+    ptys.length = 0;
+    hookCb = undefined;
+    lifecycle = new PtyLifecycleManager({ maxLivePtys: 10 });
+    const hookRegistry = {
+      register: (_id: string, cb: (h: any) => void) => { hookCb = cb; },
+      unregister: () => {},
+    } as any;
+    engine = new InteractiveClaudeEngine(lifecycle, hookRegistry);
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  /** Turn 1 cold-spawns and completes via Stop, which leaves the PTY warm so the
+   *  next turn takes the paste-and-submit path. */
+  async function completeColdTurn(sessionId: string): Promise<FakePty> {
+    const turn = engine.run({ sessionId, prompt: "first", cwd: "/tmp" } as any);
+    await vi.advanceTimersByTimeAsync(20);
+    hookCb!({ hook_event_name: "SessionStart", session_id: "c1" });
+    hookCb!({ hook_event_name: "Stop", last_assistant_message: "done" });
+    const r = await turn;
+    expect(r.error).toBeUndefined();
+    const pty = ptys[0];
+    expect(pty).toBeDefined();
+    return pty;
+  }
+
+  it("takes UserPromptSubmit as the acknowledgement, and SessionStart as nothing", async () => {
+    vi.useFakeTimers();
+    const pty = await completeColdTurn("s-ack");
+    expect(engine.hasWarmPty("s-ack")).toBe(true);
+
+    void engine.run({ sessionId: "s-ack", prompt: "second", cwd: "/tmp" } as any);
+    await vi.advanceTimersByTimeAsync(20);
+    const before = crs(pty);
+
+    // The initial submit CR, 150ms after the bracketed paste.
+    await vi.advanceTimersByTimeAsync(200);
+    expect(crs(pty)).toBe(before + 1);
+
+    // A stale SessionStart — exactly what the idle spawn that warmed this PTY
+    // emits. Accepting it would confirm a submit that never happened.
+    hookCb!({ hook_event_name: "SessionStart", session_id: "c2" });
+    await vi.advanceTimersByTimeAsync(1600);
+    expect(crs(pty)).toBe(before + 2); // still retrying
+
+    hookCb!({ hook_event_name: "UserPromptSubmit", prompt: "second" });
+    const afterAck = crs(pty);
+    await vi.advanceTimersByTimeAsync(10_000);
+    expect(crs(pty)).toBe(afterAck); // acknowledged: no further CRs
+
+    expect(engine.turnProgress("s-ack")?.awaitingSubmit).toBe(false);
+  });
+
+  it("reports awaitingSubmit only while the prompt is unacknowledged", async () => {
+    vi.useFakeTimers();
+    await completeColdTurn("s-progress");
+
+    void engine.run({ sessionId: "s-progress", prompt: "second", cwd: "/tmp" } as any);
+    await vi.advanceTimersByTimeAsync(20);
+
+    const pending = engine.turnProgress("s-progress");
+    expect(pending?.awaitingSubmit).toBe(true);
+    expect(pending?.activeTools).toBe(0);
+    expect(pending?.activeUpstream).toBe(false);
+    // lastProgressAt is the newest of turn start, last hook and last output, so
+    // it must be a real instant at or before now even before anything happens.
+    expect(pending?.lastProgressAt).toBeGreaterThan(0);
+    expect(pending?.lastProgressAt).toBeLessThanOrEqual(Date.now());
+
+    // A tool in flight is real work, so the gateway must not read this as quiet.
+    hookCb!({ hook_event_name: "PreToolUse", tool_name: "Bash", tool_input: {} });
+    expect(engine.turnProgress("s-progress")?.activeTools).toBe(1);
+    // ...and any in-turn hook proves the prompt is running.
+    expect(engine.turnProgress("s-progress")?.awaitingSubmit).toBe(false);
+
+    hookCb!({ hook_event_name: "PostToolUse", tool_name: "Bash" });
+    expect(engine.turnProgress("s-progress")?.activeTools).toBe(0);
+  });
+
+  it("does not confirm a cold spawn, which carries its prompt in argv", async () => {
+    vi.useFakeTimers();
+    const turn = engine.run({ sessionId: "s-cold", prompt: "only", cwd: "/tmp" } as any);
+    await vi.advanceTimersByTimeAsync(20);
+
+    // Submitted by construction: nothing is pasted and no CR is due.
+    expect(engine.turnProgress("s-cold")?.awaitingSubmit).toBe(false);
+    await vi.advanceTimersByTimeAsync(10_000);
+    expect(crs(ptys[0])).toBe(0);
+
+    hookCb!({ hook_event_name: "SessionStart", session_id: "c1" });
+    hookCb!({ hook_event_name: "Stop", last_assistant_message: "done" });
+    await turn;
+  });
+
+  it("does not confirm a native command, which never emits the hook", async () => {
+    vi.useFakeTimers();
+    const pty = await completeColdTurn("s-native");
+    const before = crs(pty);
+
+    void engine.run({ sessionId: "s-native", prompt: "/compact", cwd: "/tmp" } as any);
+    await vi.advanceTimersByTimeAsync(20);
+
+    // /compact runs locally and settles on its own timer. One CR submits it; a
+    // retry loop would fire spurious CRs at a prompt that already did its work.
+    expect(engine.turnProgress("s-native")?.awaitingSubmit).toBe(false);
+    await vi.advanceTimersByTimeAsync(200);
+    expect(crs(pty)).toBe(before + 1);
+    await vi.advanceTimersByTimeAsync(10_000);
+    expect(crs(pty)).toBe(before + 1);
+  });
+
+  it("writes no further CRs once the turn settles unacknowledged", async () => {
+    vi.useFakeTimers();
+    const pty = await completeColdTurn("s-cancel");
+
+    const turn = engine.run({ sessionId: "s-cancel", prompt: "second", cwd: "/tmp" } as any);
+    await vi.advanceTimersByTimeAsync(200); // initial CR sent, prompt unacknowledged
+    const before = crs(pty);
+
+    // Settled without ever seeing UserPromptSubmit — the case where the loop
+    // would otherwise outlive its turn and submit whatever the composer holds
+    // when the warm PTY is handed to the next one.
+    engine.kill("s-cancel", "Interrupted: new message received");
+    await turn;
+
+    await vi.advanceTimersByTimeAsync(30_000);
+    expect(crs(pty)).toBe(before);
+    expect(engine.turnProgress("s-cancel")).toBeUndefined();
+  });
+
+  // NOT tested here: that run()'s finally calls entry.cancelSubmitConfirm().
+  // `submitted()` also reads resolver.isSettled, so a settled turn stops the loop
+  // at its next tick with or without that call — removing it changes no
+  // observable behaviour, confirmed by mutation. The only difference is that the
+  // interval lives ~1.5s longer, and timer counts cannot isolate it from the
+  // per-session timers armed alongside. The cancel stays as redundancy; the
+  // guarantee that matters is pinned by the test above.
+});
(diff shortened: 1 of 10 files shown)


hristo2612:
Reviewed. **MERGE-AFTER-CHANGES** — this targets a hole that is genuinely still open on v0.28.6, and the acknowledgement mechanism you chose is the right one. Six changes needed, three of them substantive.

Context you may not have: v0.28.6 (`88ddd669`) fixed the attachment hang by backticking attachment paths, after proving on a live PTY that the trigger is the TUI's **image auto-attach** (an async encode that discards keypresses, swallowing the 150 ms submit CR) rather than newlines. A prior PR (#78) proposing a 20 s watchdog + PTY kill + cold-resume was rejected.

## What you're fixing that main doesn't — and it's real

`buildAttachmentSuffix` backticks only the "Attached files:" list. So a bare path to a real image in the **message body** still triggers the same auto-attach race, and on main today that turn hangs for the full 15 minutes before `shouldSettleStalledTurn` settles it as an error. Your fix would rescue it — and per my own measurements a CR re-sent at +1650 ms *succeeds* where +150 ms fails, so this is a real improvement, not just faster failure.

(For the record: `writeStdin` at `claude-interactive.ts:1249` is still broken on main too — I incorrectly claimed otherwise on #78 and have corrected that. You explicitly scoped it out; noted, and it needs its own fix either way.)

Your rebase is clean — the branch is literally main + 1 commit, zero conflicts, and it doesn't touch `buildAttachmentSuffix` or `shouldSettleStalledTurn`. Suites green with it applied (jinn-cli 3494, web 1140, typecheck clean both).

**The confirmation mechanism is sound.** Using Claude Code's `UserPromptSubmit` hook — rather than sniffing PTY output or a quiet-detector — is exactly right, and excluding `SessionStart` because the idle spawn fires it shows you traced it properly.

## Required changes

**1. Gate `onUnconfirmed` on `hasActiveUpstream()`.** This is the specific flaw I rejected #78 for. Right now it consults only `promptSubmitted || resolver.isSettled`. Concrete failure: the gateway dispatches a turn while a **PTY-native** turn (a human typing in the CLI view) is still running; Claude Code queues your prompt and fires no `UserPromptSubmit` until it dequeues. If that human turn is pure text with no tool calls and runs >6.15 s, you settle a prompt that is alive and queued. `hasActiveUpstream()` (`claude-interactive.ts:838`) is a free map lookup — please also consider `entry.activeTools === 0`.

**2. 6.15 s is too tight** (`150 + 4×1500`). `assets/hook-relay.mjs` is a fresh `node` process per hook: cold boot + `gateway.json` read + loopback POST. On a box running several concurrent PTYs — the exact state you describe — exceeding that is plausible, and every overshoot kills a live turn. Either raise to ~20-30 s total, or better: make `onUnconfirmed` **report only** and let the existing 15-minute backstop own settling. The retry is the valuable part; the verdict is the risky part.

**3. Drop the `Interrupted:` prefix on the unconfirmed error.** It's load-bearing downstream: `manager.ts:641` / `api.ts:6916` do `error?.startsWith("Interrupted")` → quiet-preempt → `manager.ts:942` calls `notifyParentSession` with `error: null`. A delegated child turn would vanish with no error reaching the parent. The backstop comment at `claude-interactive.ts:919-922` says exactly this: *"Not prefixed 'Interrupted:' on purpose — that triggers quiet-preempt handling downstream, and a stall must surface as a real error."*

**4. Clear the composer before settling.** The comment says "the pasted text is still in the composer, so the operator loses nothing" — but if the PTY stays warm, the next dispatched message bracket-pastes on top of the stranded text and one CR submits both as a single merged prompt. Main has the same exposure via the 15-min backstop; this makes it ~150× more likely. Send `\x15`/ESC or force a cold respawn.

**5. The amber dot may never appear in the case it exists for.** `turnStall` is derived at serialize time, but nothing pushes it: the 5 s heartbeat emits no `session:updated`, the web query has `staleTime: 5 * 60_000` and no `refetchInterval`, and invalidation is WS-event-driven. A wedged session generates no events, so on a quiet instance the row never re-serializes. Also, shipping `stalledForMs` (a duration) rather than `lastProgressAt` (an instant) means the label freezes at the last refetch and never ticks. Send `lastProgressAt`, derive age client-side, and add a refresh path.

**6. Tests don't cover the risky logic.** All six new engine tests drive `pasteAndSubmit` with a hand-stubbed `submitted()` — one even says so (`// stands in for 'promptSubmitted || resolver.isSettled'`). Nothing tests that a `UserPromptSubmit` hook actually sets `promptSubmitted`, that `promptSubmitted: !warm || nativeCommand` is right, or that `entry.cancelSubmitConfirm?.()` runs in the `finally`. `computeTurnStall()` and `turnProgress()` have no tests at all, and `computeTurnStall` is easily testable. Also: the new tests were appended inside `describe("shouldSettleStalledTurn")` while a `describe("pasteAndSubmit")` already exists at line 192.

## Smaller notes

- `HookRegistry` buffers with a 30 s TTL and drains into the next `register()`, so a hook from a prior PTY-native turn can falsely ack. Fails safe (no retry), but weakens the fix on busy CLI-view sessions.
- Registering the relay on `UserPromptSubmit` puts a synchronous Node spawn on the critical path of **every** turn (~100-200 ms) and widens the blast radius of the relay's exit-2 branch, where a non-zero exit hard-blocks and erases the prompt.

Address 1-4 and I'll take it; 5-6 I'd like too but they're not blockers for correctness. Good instinct going after the acknowledgement signal rather than a timer.

e1010101:
All six addressed. The shape changed for 1–4, so worth reading together.

**1. Gated on `hasActiveUpstream()`** — and applied to the **retry as well as the verdict**, since spraying CRs at a busy TUI is its own hazard, not just the false report. The gate is `hasActiveUpstream(id) || entry.activeTools > 0`, as you suggested. It holds the loop open without spending an attempt.

**2–4 resolved together by making `onUnconfirmed` report-only**, which was your preferred option and I agree with the reasoning: the retry is the value, the verdict is the risk. That moots 3 and 4 rather than fixing them separately — no error string means nothing reaches the `startsWith("Interrupted")` quiet-preempt path, and not settling early keeps the composer-merge exposure at the backstop's existing rate instead of ~150× more likely. `shouldSettleStalledTurn` now owns the verdict outright.

The cost is real and I want to name it: losing a CR now costs the 15-minute window instead of six seconds. I raised the retry budget from 3 to 12 to compensate, because by your own measurement a CR re-sent at +1650 ms *succeeds* where +150 ms fails — so most swallowed CRs should be recovered by a retry and never reach the verdict at all.

**5. Now sends `lastProgressAt`**, an instant, with the age derived client-side; there is a test that the label keeps advancing across an hour without a refetch. I did **not** add the push/refresh path — that is real work in the query layer and I would rather not bundle it into this PR. So on a quiet instance the amber row can still be late to appear. Say the word if you want it here rather than as a follow-up.

**6. Partly.** Added the `busy` gate and report-only tests, and moved the submit cases out of `describe("shouldSettleStalledTurn")` into `describe("pasteAndSubmit")` — you were right that I had appended them into the wrong block. Still missing the ones that need engine-level harnessing rather than a stub: that a real `UserPromptSubmit` hook sets `promptSubmitted`, the `!warm || nativeCommand` initialiser, `cancelSubmitConfirm` running in the `finally`, and `computeTurnStall`. I would rather flag those as outstanding than fake them with more stubs.

On your smaller notes: the `HookRegistry` 30s buffer draining into the next `register()` is a real weakening on busy CLI-view sessions and I have not fixed it — with report-only it now fails toward "no retry" rather than a bad verdict, which is the safer direction. And I take the point about putting a Node spawn on every turn's critical path via `UserPromptSubmit`; that is the cost of using the acknowledgement rather than a timer, and I do not have a cheaper signal.

Thanks for the #78 context — knowing a watchdog-plus-kill had already been rejected is what made report-only the obvious answer rather than a concession.

e1010101:
Point 5 is now fully addressed in code, not deferred.

**The problem, precisely.** `turnStall` was computed in `serializeSession`, so the client only learned a session was stalled when it refetched `/api/sessions`. The sessions query has `staleTime: 5 * 60_000`, no `refetchInterval`, and `refetchOnMount: false`; invalidation is driven by WS events. A wedged session emits no events — no hooks, no stream deltas — so nothing invalidated the query. The amber row therefore depended on the very activity whose absence it existed to report, and would only appear if something unrelated happened to trigger a refetch.

**The fix.** The verdict moved to the client:

- `computeLiveTurnProgress` (was `computeTurnStall`) reports `lastProgressAt` for *any* live turn and passes no staleness judgement. The client already receives a serialize at turn start, while the session is still healthy, so it holds the instant before the silence begins.
- `TURN_STALL_VISIBLE_MS = 90_000` now lives in `chat-sidebar.tsx`, applied inside `getTurnStall` against a `now` argument.
- `useStallClock()` is a single `useSyncExternalStore` interval (15s) shared by the whole sidebar rather than a timer per row. Rows with nothing in flight pass `active: false` and never subscribe, so a quiet list starts no timer at all; the snapshot is refreshed on subscribe so a remount after a long idle does not read a stale cached instant.

The row now crosses the threshold on its own clock with zero network traffic, which is the case the feature exists for.

**What stayed on the server.** `turnProgress` is still null while a tool or upstream request is in flight. That is state rather than time, so the gateway is the right place to judge it, and both edges of it emit hooks — hence events, hence a refetch — so the clearing signal is guaranteed.

**Tests.** `session-serialization.test.ts` gains coverage for `turnProgress`: reported 5s into a turn (no server-side threshold), `awaitingSubmit` passed through, null while `activeTools > 0` or `activeUpstream`, null for non-running sessions and for engines that cannot report progress. `chat-sidebar-helpers.test.ts` pins that the threshold is applied client-side, that one payload from turn start crosses it unaided and keeps advancing across an hour, and that `useStallClock` ticks, tears its interval down on unmount, and starts no timer when inactive.

Full `packages/web` suite and the `packages/jinn` gateway/engine suites pass, except for failures that are pre-existing on this branch (Windows-only; they are what #97, #100 and #101 address).

e1010101:
Point 6 is now actually done — my earlier attempt did not land. The nine submit tests were still appended inside `describe("shouldSettleStalledTurn")`; they now sit under `describe("pasteAndSubmit")` with no assertion changed.

Also closed one of the gaps I listed as outstanding. `SUBMIT_ACK_HOOKS` is exported and pinned by a test, because the `SessionStart` exclusion is the load-bearing part and nothing guarded it: the idle spawn that warms a PTY fires `SessionStart` before the paste, so accepting it would confirm a submit that never happened and leave the prompt sitting in the composer — precisely the failure the confirmation exists to catch.

The two remaining gaps from that list — that `promptSubmitted` initialises to `!warm || nativeCommand`, and that `cancelSubmitConfirm` runs in `run()`'s `finally` — are wiring inside `run()`, which needs a live PTY, hook registry and lifecycle to exercise. The decisions they encode are covered where they live: `cancel()` stopping the retry loop is pinned by `cancel() stops the retry loop, so a settled turn cannot type into the next one` and `cancel() before the initial CR suppresses the submit entirely`. I would rather leave those two honestly untested than build a mock deep enough to assert them, but say the word and I will.

hristo2612:
Re-verified at head `a1acde87` against current `main`. **The code is approved** — what's blocking is the PR description, plus two test gaps I'd like to land with it.

Verified green: jinn-cli **283 files / 3503 passed | 1 skipped**, web **107 / 1145**, typecheck clean, merges cleanly onto `main`.

## Credit first — the five addressed points hold up

The `busy` gate (`claude-interactive.ts:983`, `hasActiveUpstream || activeTools > 0`) is the right answer to the queued-vs-swallowed ambiguity, and applying it to the retry as well as the verdict was a better call than what I asked for. `SUBMIT_ACK_HOOKS` correctly excludes `SessionStart` with a comment explaining the idle-spawn trap. The cancellation contract is genuinely load-bearing and both the `finally` and the pre-CR case are covered. And moving the stall verdict client-side is the correct architecture — a wedged session emits no events, so a server-computed verdict would never be delivered; I traced the delivery path end to end and the amber row does render at 90s with no further server traffic.

Making `onUnconfirmed` report-only was the right resolution for points 2-4. It doesn't just dodge them — no error string means the `startsWith("Interrupted")` quiet-preempt path is unreachable, and not settling early keeps the composer-merge exposure at the backstop's existing rate rather than ~150× higher.

## Blocking: the description contradicts the code

This text becomes the merge-commit record and is the first thing anyone reads when this behaviour is next debugged. Four claims are wrong at `a1acde87`:

| Body says | Code does |
|---|---|
| "if it never does the turn settles as interrupted rather than hanging" | Report-only — `onUnconfirmed` (`:995-998`) emits a `logger.warn` and nothing else. Your own comment at `:990-994` explains why settling there would be wrong. |
| "`turnIsTerminallyStalled` already does that" | The function is `shouldSettleStalledTurn` (`:545`). No `turnIsTerminallyStalled` exists in the tree. |
| "`turnStall` is derived at serialize time" | The serialized field is `turnProgress` (`api.ts:2014`); `getTurnStall` is a **client** helper. Your own commit `05ec365e` is titled "Move the stall verdict to the client", so this is doubly stale. |
| "Tests cover … the exemptions" | They don't — see below. |

The first row is the one that matters: a reader will believe an unconfirmed submit terminates the turn, and go looking for the bug in the wrong place when it doesn't. It's a text edit and costs nothing.

## Should land with it: the wiring is untested

`grep -rn "promptSubmitted" packages/jinn/src` returns no test file. The mechanism is well covered at the `pasteAndSubmit` level, but none of these is exercised through `engine.run`:

1. A `UserPromptSubmit` via `hookCb` sets `promptSubmitted` and stops the retry loop, while `SessionStart` does **not** (`:952-956`). Currently only `SUBMIT_ACK_HOOKS.has(...)` is unit-tested — the wiring from the hook callback to `entry.promptSubmitted` isn't.
2. `promptSubmitted: !warm || nativeCommand` (`:935`) and the `nativeCommand ? undefined : {…}` argument at `:975` — the two exemptions the description claims are tested.
3. The `finally` calling `entry.cancelSubmitConfirm?.()` (`:1093`).

This is cheaper than it sounds: `claude-interactive-race.test.ts` already captures `hookCb` and drives `engine.run`, and a turn 1 that completes via `Stop` leaves a **warm** PTY so `injectPrompt` is reachable. The only harness change needed is making `FakePty.write` record instead of no-op. Concretely: turn 1 cold spawn → `Stop` → turn 2 warm with fake timers → assert a CR retry fires → feed `SessionStart`, assert it still retries → feed `UserPromptSubmit`, assert retries stop. Same harness with `prompt: "/compact"` covers the native-command exemption, and `engine.kill()` mid-turn covers the `finally`.

Separately, `InteractiveClaudeEngine.turnProgress()` itself is untested — `session-serialization.test.ts:62` asserts against a hand-written `{ turnProgress: () => progress }` stub, so `Math.max(startedAt, lastHookAt, lastOutputAt)` and `awaitingSubmit: !promptSubmitted` are unexercised. The same warm-PTY harness makes this trivial.

## Non-blocking

`TurnProgress.turnStartedAt` is produced (`:1463`) and declared (`types.ts:166`) but never consumed — `computeLiveTurnProgress` returns only `lastProgressAt` and `awaitingSubmit`, and nothing in either package reads it. Either wire it through (turn-elapsed would strengthen the "no output for 51m" chip) or drop it; a struct member with no reader is a false affordance for the next caller.

## Also worth knowing

`#97`, `#96`, `#100` and `#95` just landed on `main`. None touches `engines/claude-interactive.ts`, so you shouldn't see conflicts, but a rebase before the final push is cheap insurance.

Fix the description and add the wiring tests and I'll merge this. The implementation itself I'm happy with.

e1010101:
Both items done, rebased onto current `main` (4 commits, no conflicts).

## The description is corrected

All four claims you listed are fixed, and the first one properly rather than by softening it — the body now says `onUnconfirmed` logs and nothing else, and gives the three reasons that is the right resolution (queued-vs-swallowed is undecidable from acknowledgement alone, `shouldSettleStalledTurn` owns the verdict, and no error string keeps the `startsWith("Interrupted")` path unreachable). `turnIsTerminallyStalled` is gone, the client/server split is described as it is built, and the test paragraph now lists what exists.

## The wiring is tested

`claude-interactive-submit-wiring.test.ts`, built on the harness you pointed at — `claude-interactive-race.test.ts`'s captured `hookCb`, with `FakePty.write` recording instead of no-op, and turn 1 completing via `Stop` to leave the PTY warm. Five tests:

1. `UserPromptSubmit` on the hook callback sets `promptSubmitted` and stops the retry loop; a `SessionStart` fed in between does **not** — the CR count keeps climbing across it and stops only on the real ack.
2. `turnProgress()` from the real engine: `awaitingSubmit` tracks the acknowledgement, `activeTools` follows `PreToolUse`/`PostToolUse`, `lastProgressAt` is a real instant. No hand-written stub.
3. The cold-spawn exemption: `awaitingSubmit` false immediately and **zero** CRs over 10s.
4. The native-command exemption: `/compact` submits once and never retries.
5. A turn settled unacknowledged writes no further CRs.

**I mutation-checked each one** rather than trusting that they bite. Adding `"SessionStart"` to `SUBMIT_ACK_HOOKS` fails (1); changing `promptSubmitted: !warm || nativeCommand` to `!warm` fails (4). Clean tree: 5 passed.

## One thing I could not honestly test, and what I found

Your third item — the `finally` calling `entry.cancelSubmitConfirm?.()`. I wrote a test for it, mutation-checked the test, and **it did not bite**: removing the call left all tests green. The reason is that `submitted()` also reads `resolver.isSettled`, so any settle stops the loop at its next tick regardless. The only difference the `finally` makes is that the interval is gone immediately rather than up to 1.5s later.

I then tried to pin that timing property via `vi.getTimerCount()` — twice — and both attempts were confounded by per-session timers armed alongside (the second measured 2 vs a reference of 1 for reasons unrelated to the cancel). Rather than ship a test that looks like coverage but measures something else, I removed it and wrote the finding into the test file next to the tests that do bite. The cancel stays as redundancy. If you want it pinned anyway, the only instrument left is spying on `clearInterval`, which asserts implementation rather than behaviour — happy to add that if you'd prefer it there.

## Non-blocking item

`TurnProgress.turnStartedAt` is dropped. You were right that nothing read it — I confirmed across both packages — and a struct member with no reader is a false affordance. Wiring turn-elapsed into the chip would be a UI change I did not want to fold into an approved PR; if you want it, it is easy to add back with a consumer.

## Verified

jinn-cli typecheck clean, web typecheck clean, `src/engines` + `session-serialization` 411 passed, web sidebar suite 40 passed. The full Linux CI path (`install --frozen-lockfile` → `typecheck` → `build` → `test`) was reproduced in a `node:24.13.0` container at the previous head with upstream `main` as a control: both green, 283 + 107 test files. The only failure on my Windows box is `pty-stream.test.ts`'s 30s timeout, which reproduces identically on `main`.

e1010101:
Rebased onto `main` at `ba4cd3df` (v0.28.7) — five commits, no conflicts. Nothing in that release touches `engines/claude-interactive.ts`, as you predicted.

Verified after the rebase: jinn typecheck clean, web typecheck clean, `claude-interactive-submit-wiring` + `claude-interactive` + `session-serialization` 55 passed, and the full web suite 107 files / 1146 passed.

PR #95, opened 2026-07-26
fix(migrations): let a no-op migration complete without symlink permission

e1010101:
A patch upgrade that changes no instance files cannot complete on Windows without symlink privilege, and the dashboard reports it as "the migration service is temporarily unavailable" — which reads as a transient outage for a condition that never clears on its own. Both halves are fixed here: the snapshot no longer materializes symlinks at all, and a failure now says what actually went wrong.

---

**Who this reaches.** `jinn setup` has had a symlink→copy fallback for `AGENTS.md` since v0.19.0 (`cli/setup.ts:551-570`), so an instance created *without* the privilege has a regular file and never reaches `fs.symlinkSync`. The affected layout is an instance whose `AGENTS.md` **is** a symlink and which now runs without the privilege: created with Developer Mode on and later off, elevated at setup but not at runtime, or a `~/.jinn` synced from a POSIX machine. Narrower than "every Windows instance" — but not hypothetical: the instance this was found on has the symlink, and its gateway log carries the `EPERM ... symlink` twice, once per upgrade attempt.

**An empty bundle still demanded a snapshot.** `selectedPaths()` unconditionally added `config.yaml`, `CLAUDE.md` and `AGENTS.md` regardless of whether the bundle carried any changed records. With no changed records there is no three-way comparison for those context files to support, so it now returns nothing. Creation and verification both derive their file set from that one function, so they stay consistent by construction, and the snapshot directory and `snapshot.json` are still written — which is what the completion receipt and its gate require.

That alone would have unblocked patch upgrades and left feature upgrades broken on the same machine (0.26.0 ships 16 files, 0.28.0 ships 4), so the general fix is folded in: **`copyEntry` no longer materializes symlinks.** The receipt already records `linkTarget` and its sha256, which is the entire content of a symlink, and no code path restores a snapshot from disk — the snapshot is a verification artifact plus a home for the completion receipt. `verifyMigrationSnapshot` checks the link from the receipt and accepts either its absence or a link left by an older build, so snapshots written before this change keep verifying rather than throwing `existing migration snapshot failed verification` on the next open. A bundle that does carry records keeps all of its comparison material; that is pinned by a test.

**The failure said nothing useful.** Every error from `openInstanceMigration` collapsed into one opaque sentence, so a fixable local condition looked like a service outage with nothing pointing at the remedy. The route now attaches a `remedy` naming Developer Mode when the platform refused a symlink, and the dialog shows it in place of the generic message. The raw error stays in the log and out of the body, per the convention the rest of `api.ts` follows.

Note that with `copyEntry` no longer creating symlinks, that classifier is a **backstop** rather than the live path for this bug — the snapshot itself no longer asks for the privilege. It is kept because the message is the difference between a fixable condition and an apparent outage for any other symlink-dependent failure in this route.

**Verification.** An empty required bundle now opens end to end through the API over a symlinked `AGENTS.md`: 201, snapshot receipt with `"files": []`, nothing materialized under it, the instance's own symlink untouched. The snapshot suite pins that a symlink is recorded but not recreated, that a pre-existing on-disk link still verifies, and that one disagreeing with the receipt does not.

Note for reviewers on Windows: `snapshot.test.ts` has one pre-existing failure here (`expected 438 to be 416`), a POSIX permission assertion Windows cannot satisfy. It fails on `main` as well and is addressed in #97.

### DIFF
--- a/packages/jinn/src/gateway/__tests__/instance-migration-api.test.ts
+++ b/packages/jinn/src/gateway/__tests__/instance-migration-api.test.ts
@@ -44,6 +44,24 @@ function seedBundle() {
   fs.writeFileSync(path.join(dir, "MIGRATION.md"), "# migration\n")
 }
 
+/** The shape every patch release actually ships: a required chain bridge that
+ *  changes no files. `AGENTS.md` is a symlink here because that is what
+ *  `jinn setup` creates, and it used to drag the snapshot through
+ *  fs.symlinkSync. */
+function seedEmptyBundle() {
+  seedBundle()
+  fs.symlinkSync("CLAUDE.md", path.join(home, "AGENTS.md"))
+  const dir = path.join(migrationsDir, "0.26.0")
+  fs.rmSync(path.join(dir, "files"), { recursive: true, force: true })
+  fs.writeFileSync(path.join(dir, "manifest.json"), JSON.stringify({
+    schemaVersion: 1,
+    version: "0.26.0",
+    baseVersion: "0.25.0",
+    generatedFrom: { baseRef: "v0.25.0", headRef: "WORKTREE" },
+    files: [],
+  }, null, 2) + "\n")
+}
+
 function responseCapture() {
   let status = 200
   const chunks: Buffer[] = []
@@ -198,6 +216,32 @@ describe("instance migration API", () => {
     fs.writeFileSync(path.join(migrationsDir, "0.26.0/files/target/CLAUDE.md"), "tampered")
     expect((await request("GET", "/api/instance-migration")).status).toBe(500)
     expect(registry.listSessions()).toHaveLength(0)
+  })
+
+  it("opens an empty required bundle end to end, over a symlinked AGENTS.md", async () => {
+    // The case that was dead on Windows: 0.27.0 and every 0.28.x ship
+    // "files": [], and the snapshot dragged AGENTS.md through fs.symlinkSync
+    // regardless, so the open route 500'd as "temporarily unavailable" and the
+    // instance could never leave the migration gate.
+    seedEmptyBundle()
+    const pending = await request("GET", "/api/instance-migration")
+    expect(pending.body).toMatchObject({ required: true, fromVersion: "0.25.0", toVersion: "0.26.0" })
+    expect(pending.body.changedFiles ?? []).toEqual([])
+
+    const opened = await request("POST", "/api/instance-migration/open", { migrationKey: pending.body.migrationKey })
+
+    expect(opened.status).toBe(201)
+    expect(opened.body).toMatchObject({ reused: false, migrationKey: pending.body.migrationKey })
+    expect(opened.body).not.toHaveProperty("remedy")
+    expect(dispatched).toHaveLength(1)
+
+    // The snapshot still exists and holds the receipt the completion gate reads,
+    // and nothing was materialized under it.
+    const snapshotDir = path.join(home, ".migration-snapshots", pending.body.migrationKey)
+    expect(JSON.parse(fs.readFileSync(path.join(snapshotDir, "snapshot.json"), "utf8")).files).toEqual([])
+    expect(fs.existsSync(path.join(snapshotDir, "AGENTS.md"))).toBe(false)
+    // The instance's own symlink is untouched.
+    expect(fs.lstatSync(path.join(home, "AGENTS.md")).isSymbolicLink()).toBe(true)
   })
 
   it("classifies a refused symlink so the operator is told which knob to turn", () => {
--- a/packages/jinn/src/gateway/__tests__/instance-migration-api.test.ts
+++ b/packages/jinn/src/gateway/__tests__/instance-migration-api.test.ts
@@ -199,4 +199,24 @@ describe("instance migration API", () => {
     expect((await request("GET", "/api/instance-migration")).status).toBe(500)
     expect(registry.listSessions()).toHaveLength(0)
   })
+
+  it("classifies a refused symlink so the operator is told which knob to turn", () => {
+    // Windows without SeCreateSymbolicLinkPrivilege reports EPERM from
+    // fs.symlinkSync. Collapsing that into a bare "service unavailable" hid a
+    // fixable local condition behind a message that reads as a transient outage.
+    const denied = Object.assign(new Error("EPERM: operation not permitted, symlink 'CLAUDE.md' -> '/snap/AGENTS.md'"), {
+      code: "EPERM",
+      syscall: "symlink",
+    })
+    expect(api.isSymlinkPrivilegeError(denied)).toBe(true)
+    expect(api.isSymlinkPrivilegeError(Object.assign(new Error("nope"), { code: "EACCES", syscall: "symlink" }))).toBe(true)
+    // Message-only fallback, for errors that lost their syscall metadata.
+    expect(api.isSymlinkPrivilegeError(new Error("EPERM: operation not permitted, symlink 'a' -> 'b'"))).toBe(true)
+
+    // Unrelated failures must NOT claim a symlink remedy.
+    expect(api.isSymlinkPrivilegeError(Object.assign(new Error("ENOSPC: no space"), { code: "ENOSPC", syscall: "write" }))).toBe(false)
+    expect(api.isSymlinkPrivilegeError(new Error("existing migration snapshot failed verification"))).toBe(false)
+    expect(api.isSymlinkPrivilegeError("EPERM symlink")).toBe(false)
+    expect(api.isSymlinkPrivilegeError(undefined)).toBe(false)
+  })
 })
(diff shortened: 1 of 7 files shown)


hristo2612:
Reviewed. Rebases cleanly, typecheck clean, suites green with it applied. **MERGE-AFTER-CHANGES** — the mechanism is sound, but the stated root cause looks wrong and the fix is narrower than it appears.

## Verified as correct

- **Empty bundles are the normal case, not an edge case.** Every shipped bundle from `0.27.0` and `0.28.1`–`0.28.6` has `"files": []` — these are deliberate "empty chain bridges" that keep the strict migration chain aligned. So `changedFiles === []` with `required: true` is the standard patch-upgrade shape, and it's worth handling well.
- **The bypass doesn't skip work for a real migration.** `createMigrationSnapshot` and `verifyMigrationSnapshot` both derive their set from `selectedPaths()`, so they stay in lockstep, and `completeInstanceMigration` gates on a receipt covering every `changedFiles` entry — vacuously satisfied when empty, which is correct. Non-empty bundles are untouched and you pinned that with a test. Good.

## The root cause claim is probably wrong

`cli/setup.ts:551-570` has had a symlink→copy fallback for `AGENTS.md` since v0.19.0, and before that it was a plain file. So on a Windows instance created by `jinn setup` **without** symlink privilege, `AGENTS.md` is a regular file and `copyEntry` never calls `symlinkSync` at all. Reaching that EPERM requires Developer Mode ON at setup and OFF later, or a home synced from a POSIX machine.

Your PR body says this was verified against a *simulated* machine — that's consistent with forcing a symlink rather than reproducing the real Windows layout. Please either correct the description to state the narrow reachability, or show a real repro. (There's no filed issue for it; #98 and #99 are unrelated.)

## The fix defers the actual problem

For anyone who *does* have a symlinked `AGENTS.md`, the next release with a **non-empty** bundle (0.26.0 had 16 files, 0.28.0 had 4) hits the identical `fs.symlinkSync` EPERM at `snapshot.ts:104` — a path this PR deliberately leaves unchanged. So it unblocks patch upgrades and leaves feature upgrades broken on the same machine.

The general fix is small and in the same file: **`copyEntry` doesn't need to materialize symlinks at all.** `snapshot.json` already records `linkTarget` and its sha256, so the on-disk symlink carries no extra information — skip materialization for `state === "symlink"` and relax the matching branch in `verifyMigrationSnapshot`. That covers empty *and* non-empty bundles. Either fold that in or file it as a follow-up, but please don't leave it implicit.

**Note for cross-referencing #98:** a shared try-symlink→hardlink→copy helper is right for #98, `setup.ts:636` and `watcher.ts:63`, but it does **not** work here — `verifyMigrationSnapshot` asserts `stat.isSymbolicLink()`, so a hardlink or copy would fail the snapshot's own verification. Genuinely two different problems.

## Smaller items

1. **Drop `detail` from the 500 body.** It carries raw error text including absolute instance paths, and it breaks this file's own convention — every other 500 in `api.ts` logs the message and returns a generic body (see `MIGRATION_BUNDLE_INVALID` two blocks up). Nothing consumes it either: the web gate reads only `remedy`. Keep `remedy`, drop `detail`.
2. `remedy` is Windows-worded but returned on every platform — a macOS EPERM would advise enabling Windows Developer Mode. Minor, and it's hedged, but worth a platform check.
3. **Test gap:** the new API test only unit-tests `isSymlinkPrivilegeError`. Nothing asserts the route actually attaches `remedy`, and no test opens an empty bundle end-to-end through the API. One test that an empty-bundle open returns 201 would cover the change you actually care about.

## Latent issue worth knowing about

A snapshot created by a pre-patch build (3 context files) becomes unverifiable under the patched `selectedPaths()` → `createMigrationSnapshot` throws `existing migration snapshot failed verification`, with no self-heal; recovery means manually deleting `.migration-snapshots/<key>`. Practically gated because `migrationKey` hashes `toVersion` so a normal version bump mints a new key — but it would bite on a same-version reinstall or a backport. (Inferred from code, not reproduced.)

The `remedy` surfacing is a genuine UX improvement over "temporarily unavailable" — thanks for that part especially.

e1010101:
All five points addressed; the PR body is rewritten to match.

**Root cause claim — corrected, not defended.** You're right that `setup.ts:551-570` has had the symlink→copy fallback since v0.19.0, so an instance created without the privilege never reaches `fs.symlinkSync`. The body now states the narrow reachability explicitly: an `AGENTS.md` that *is* a symlink on a machine now running without the privilege — Developer Mode on at setup and off later, elevated setup with unelevated runtime, or a `~/.jinn` synced from POSIX.

The repro was real, though, and I should have said so plainly the first time instead of "simulated machine": the instance this was found on has `AGENTS.md` as a symlink, and its gateway log carries `EPERM: operation not permitted, symlink` twice, once per upgrade attempt. That is what started this.

**The general fix is folded in, not filed.** `copyEntry` no longer materializes symlinks at all. Your reasoning holds and goes a little further than stated: nothing anywhere restores a snapshot from disk — the snapshot is a verification artifact plus a home for `completion-receipt.json` — so `linkTarget` + its sha256 in the receipt is strictly more than the on-disk link carried.

`verifyMigrationSnapshot` verifies the link from the receipt and accepts **either** its absence **or** a link left by an older build. That second half matters for the latent issue you flagged at the end: without it, every snapshot written before this change would fail verification and `createMigrationSnapshot` would throw `existing migration snapshot failed verification` with no self-heal. A snapshot whose on-disk link *disagrees* with the receipt is still rejected.

This changed an existing assertion — `snapshot.test.ts:135` asserted the materialized link — so that is now an explicit assertion of the opposite, with the receipt checked instead of the link.

**Note on #98:** agreed, and the constraint you identified is the reason the shared helper doesn't fit — except that after this change `verifyMigrationSnapshot` no longer asserts `stat.isSymbolicLink()` on a materialized copy, because there is no materialized copy. That removes the conflict rather than resolving it; the snapshot path simply has no link to create, so the try-symlink→hardlink→copy helper stays a #98/`setup.ts`/`watcher.ts` concern.

**Smaller items.**

1. `detail` dropped. The message still goes to `logger.error`; the body is generic, matching `MIGRATION_BUNDLE_INVALID`.
2. `remedy` gated on `process.platform === "win32"`, and the Windows-specific preamble removed from the string since it can now only be seen there.
3. Test gap closed: `opens an empty required bundle end to end, over a symlinked AGENTS.md` seeds a `"files": []` manifest with a symlinked `AGENTS.md`, asserts 201, `reused: false`, one dispatch, a `snapshot.json` with `"files": []`, nothing materialized under the snapshot, the instance's own symlink untouched, and no `remedy` on the success path.

**One thing worth flagging.** With `copyEntry` no longer creating symlinks, `isSymlinkPrivilegeError` is a **backstop** on this route rather than the live path — the snapshot no longer asks for the privilege, so the remedy will not fire for the bug that motivated it. I kept it because the message is what turns any other symlink-dependent failure here from an apparent outage into a fixable condition, and its test is honestly scoped as a classifier unit test. Happy to drop it if you'd rather not carry code without a reachable trigger.

**Suites.** `src/migrations` and `instance-migration-api.test.ts` pass. Two pre-existing Windows failures remain on this branch and are not from this change (confirmed by stashing): `snapshot.test.ts` `expected 438 to be 416` (POSIX mode), and `instance-migration-bundle.test.ts`, which is flaky here — a different subset of 3-4 of its 8 tests fails per run, on `main` as well. Both are CRLF/POSIX-mode classes that #97 addresses.

e1010101:
Merge-order note: this PR and #97 both touch `snapshot.test.ts:135-138`. #97 converts the mode assertion to a `expectPosixMode` helper; this PR replaces the line above it — which asserted the materialized symlink — with the opposite assertion. Whichever lands second takes a three-line conflict, resolved by keeping both:

```ts
expect(fs.existsSync(path.join(first.path, "AGENTS.md"))).toBe(false)
expectPosixMode(fs.statSync(path.join(first.path, "CLAUDE.md")), 0o640)
```

I merged all six open PRs into one local branch to check: they otherwise combine cleanly, typecheck is clean, and the full jinn suite is 3520 passed / 4 failed, with all 4 failures reproducing identically on `main`. Details in #97.

e1010101:
Rebased onto current `main` (was 11 commits behind, at v0.28.4) — no conflicts.

Worth stating how this was verified, since GitHub is not running Actions on fork PRs here (`gh pr checks` reports no checks on any of the six). I reproduced `.github/workflows/ci.yml` on Linux in a `node:24.13.0` container — `pnpm install --frozen-lockfile` → `typecheck` → `build` → `test` — with upstream `main` (`558a3132`) run first as a control. Main is green there, and so is this branch: typecheck clean, build clean, **283 + 107 test files passed**.

The rebase mattered for that claim: CI tests `refs/pull/N/merge`, so a green run on a stale base proves nothing about what CI would actually do. This is now measured on the real base.

Two corrections to my own method, in case the numbers look surprising: my first attempt reported failures that were artifacts, not defects — the bare `node:24` image has no `lsof` (which `gateway/lifecycle.ts` shells out to on POSIX, and which `ubuntu-latest` ships), and I had invalidated the timing-sensitive `performance.test.ts` by running two containers at once. With `lsof` installed and everything run serially, main and all six branches are green.

PR #96, opened 2026-07-26
fix(skills): stop a trailing backslash escaping the Windows arg quoting

e1010101:
An argument ending in a backslash escapes its own closing quote on the Windows `npx skills` path, so it reaches the child with a literal `"` appended and swallows the argument after it. Directory arguments end in a backslash routinely, so this is ordinary input rather than a crafted one.

---

`npx` ships as `npx.cmd`, which Node refuses to spawn without a shell (EINVAL since 18.20.2, the BatBadBut mitigation). The Windows path therefore runs through `cmd` and defends itself by rejecting arguments containing shell metacharacters, then wrapping each remaining argument in quotes.

The wrapping was a bare `` `"${arg}"` ``. Windows programs parse their command line with MSVCRT rules, where a run of backslashes immediately before a quote *escapes* it, so:

```
in "trailing\"  ->  quoted "trailing\"  ->  child argv: ["trailing\"" ...]
```

Even runs are corrupted differently — silently halved rather than unbalanced. The fix is the MSVCRT rule: double only the *final* run of backslashes, which yields the intended literal backslashes and still terminates the string. Interior backslashes are untouched, so `C:\path\to\skill` is unaffected. Verified against real child processes before and after.

**On severity — corrected from my original description.** This is argument corruption and defence-in-depth, not a working injection. Every argument still passes `WINDOWS_UNSAFE_ARG`, so no `&`, `|`, `<`, `>`, `^`, `%` or newline can reach the command line even when the quoting is unbalanced; the injected character is a `"` and nothing more. Worth fixing because `jinn skills` is invoked with LLM- and web-sourced package names, and because the even-run halving would produce baffling bug reports — but it does not reach command execution.

**Also corrected: the Windows branch was not uncovered.** `skills-windows-spawn.test.ts` has covered it since `b0783896`, the same commit that added the Windows handling, by stubbing `process.platform`. My original description was wrong about that.

**Test placement fixed.** My two new spawn assertions were gated with `it.skipIf` against the *real* platform, and CI is `ubuntu-latest`, so they never executed — the regression was unguarded by the very tests meant to guard it. They now live in `skills-windows-spawn.test.ts` alongside the existing Windows cases, where the stubbed platform means they run on Linux. The `quoteWindowsArg` unit tests are platform-independent and gained an odd run of three plus the realistic `C:\skills\mine\` directory case; previously only runs of one and two were covered.

Confirmed locally: 9 of the 10 assertions across the two files execute, the single skip being the POSIX-argv contract, which runs on CI.

### DIFF
--- a/packages/jinn/src/cli/__tests__/skills-shell-security.test.ts
+++ b/packages/jinn/src/cli/__tests__/skills-shell-security.test.ts
@@ -1,4 +1,4 @@
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 
 const spawnSync = vi.hoisted(() => vi.fn(() => ({ status: 0 })));
 
@@ -8,19 +8,12 @@ vi.mock("node:child_process", () => ({
 
 import { quoteWindowsArg, runNpxSkills } from "../skills.js";
 
-// The two platforms defend this differently and the difference is the point:
-// POSIX passes argv with no shell at all, while Windows cannot spawn npx.cmd
-// without a shell (EINVAL since Node 18.20.2, the BatBadBut mitigation) and so
-// relies on rejecting metacharacters plus correct quoting. Asserting the POSIX
-// contract on both hid the Windows path from review entirely.
-const onWindows = process.platform === "win32";
-
-beforeEach(() => {
-  spawnSync.mockClear();
-});
-
 describe("skills CLI process spawning", () => {
-  it.skipIf(onWindows)("passes user-controlled skill args as argv with shell disabled", () => {
+  // The POSIX contract. The Windows branch cannot pass argv with the shell
+  // disabled — npx.cmd is unspawnable that way — and lives in
+  // skills-windows-spawn.test.ts, which stubs process.platform so those cases
+  // execute on CI instead of skipping on a non-Windows runner.
+  it.skipIf(process.platform === "win32")("passes user-controlled skill args as argv with shell disabled", () => {
     runNpxSkills(["add", "owner/repo; touch /tmp/pwned", "-g", "-y"], "pipe");
 
     expect(spawnSync).toHaveBeenCalledWith(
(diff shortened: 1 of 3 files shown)


hristo2612:
Reviewed — **the `skills.ts` change is correct and I'd take it unchanged.** This is the closest of your four PRs to landing. Three changes needed, all mechanical, none in the fix itself.

## Verified

I implemented the reference `CommandLineToArgvW` / MSVCRT parser and ran both the old and new quoting through it:

| input | old `"${arg}"` | new `quoteWindowsArg` |
|---|---|---|
| `trailing\` | `["trailing\" NEXTARG"]` — quote injected, next arg swallowed | `["trailing\", "NEXTARG"]` ✅ |
| `two\\` (even) | `["two\"]` — **silently halved** | `["two\\"]` ✅ |
| `C:\path\` | `["C:\path\" NEXTARG"]` — broken | ✅ |
| `C:\path\to\skill` | ✅ | ✅ unchanged |

The bug is real and hits ordinary Windows directory paths. Your fix is **provably complete for this path**, not partial: only backslashes immediately preceding a quote are special under MSVCRT, and `WINDOWS_UNSAFE_ARG` already rejects `"` from every argument, so the closing quote is the only quote in play — doubling the final run is therefore sufficient for both odd and even runs. Interior backslashes are untouched, and `\` isn't a cmd.exe metacharacter, so it survives the `cmd /d /s /c` layer and is resolved by npx's own parse. Full suite green with it applied (3490 passed).

## Please fix before merge

**1. Your two new spawn tests never run.** They use `it.skipIf(!onWindows)` against the *real* platform, and CI is `ubuntu-latest` only — I confirmed `2 skipped` locally. So the regression they're meant to guard is unguarded.

**2. The PR description says the Windows branch "had no coverage at all" — it does.** `packages/jinn/src/cli/__tests__/skills-windows-spawn.test.ts` was added in the same commit as the Windows handling (`b0783896`) and covers it by **stubbing** `process.platform = "win32"`. That's the correct home for your trailing-backslash spawn assertion, and stubbing is why it actually executes on Linux. Please move the two tests there rather than adding a platform-gated file.

**3. Add odd-run ≥3 and a `C:\path\` case** to the `quoteWindowsArg` unit tests. Those are platform-independent so they run everywhere — right now only 1 and 2 backslashes are covered.

## One correction to the framing

I'd soften the security wording. The old bug injects a literal `"` and merges adjacent arguments, but it **cannot reach command execution on its own**: every argument still passes `WINDOWS_UNSAFE_ARG`, so no `&`, `|`, `<`, `>`, `^`, `%` or newline can appear on the line even when quoting is unbalanced. It's argument corruption plus defense-in-depth, not a working injection. Worth having — `jinn skills` is invoked by agents with LLM- and web-sourced package names — but let's describe it accurately.

Do those and I'll merge. Thanks for finding this; the even-run halving in particular is the kind of thing that would have produced baffling bug reports.

e1010101:
All three addressed, plus the framing corrections.

**1. Tests now execute.** Moved both spawn assertions into `skills-windows-spawn.test.ts`, which stubs `process.platform` — so they run on `ubuntu-latest` instead of skipping. You were right that `it.skipIf` against the real platform left the regression unguarded by the very tests written to guard it.

**2. Description corrected** on both counts: the Windows branch was covered by `skills-windows-spawn.test.ts` since `b0783896`, and I've replaced the injection framing with argument-corruption-plus-defence-in-depth, noting explicitly that `WINDOWS_UNSAFE_ARG` keeps every shell metacharacter off the line so this cannot reach command execution.

**3. Unit tests extended** to an odd run of three, an even run of four, the realistic directory arguments `C:\path\` and `C:\skills\mine\`, and the empty string. Only runs of one and two were covered before. These are platform-independent so they run everywhere.

Confirmed locally: 9 of the 10 assertions across the two files execute, the single skip being the POSIX-argv contract, which runs on CI.

Thanks for building the reference `CommandLineToArgvW` parser — the even-run halving is the part I'd underweighted, and your table makes the completeness argument better than mine did.

e1010101:
Rebased onto current `main` (was 11 commits behind, at v0.28.4) — no conflicts.

Worth stating how this was verified, since GitHub is not running Actions on fork PRs here (`gh pr checks` reports no checks on any of the six). I reproduced `.github/workflows/ci.yml` on Linux in a `node:24.13.0` container — `pnpm install --frozen-lockfile` → `typecheck` → `build` → `test` — with upstream `main` (`558a3132`) run first as a control. Main is green there, and so is this branch: typecheck clean, build clean, **283 + 107 test files passed**.

The rebase mattered for that claim: CI tests `refs/pull/N/merge`, so a green run on a stale base proves nothing about what CI would actually do. This is now measured on the real base.

Two corrections to my own method, in case the numbers look surprising: my first attempt reported failures that were artifacts, not defects — the bare `node:24` image has no `lsof` (which `gateway/lifecycle.ts` shells out to on POSIX, and which `ubuntu-latest` ships), and I had invalidated the timing-sensitive `performance.test.ts` by running two containers at once. With `lsof` installed and everything run serially, main and all six branches are green.

PR #97, opened 2026-07-26
fix(windows): make the repo buildable, and narrow tests to what each platform can express

e1010101:
`pnpm build` fails on Windows, so the package cannot be compiled from source there at all, and the test suite silently drops about 180 tests while still reporting green.

> **Split, as requested.** This PR is now ① only. ② is #100 (port detection) and ③ is #101 (owner-only ACLs), both separate. Your `tsconfig.build.json` nit is fixed here — `test-support` was emitting into `dist/` and shipping in the tarball; `**/test-support/**` is now excluded and I verified `dist/src/shared/test-support` is gone after a build.

---

**The build.** The `jinn-cli` build script chained `rm -rf`, `mkdir -p` and `cp`. Package managers run scripts through the platform shell — `cmd` on Windows — where none of those exist, so it died with "The syntax of the command is incorrect". `scripts/build.mjs` does the same work with Node's `fs`, resolving TypeScript's own entry rather than the `tsc` shim so no shell is involved at all. Both `clean` scripts had the same flaw.

**Line endings.** Shipped migration payloads are content-addressed, and there was no `.gitattributes`, so with `core.autocrlf=true` every payload checked out CRLF while its manifest hash covers LF — the migration service failing to validate its own bundle. The blobs were always LF; only the checkout was wrong, so this needs no content change.

This one bit for real while the PR was open, which is worth recording: after pulling v0.28.6, my local `0.28.5`/`0.28.6` bundles were CRLF and the CLI and gateway computed *different migration keys* for the same upgrade. Re-checking those files out with the attributes in place made them byte-identical to the published package. Note the limitation that implies — `.gitattributes` fixes fresh clones, but files arriving later via `git fetch` onto a branch that does not carry it still land CRLF, so an existing checkout needs one renormalisation pass.

**Grok transcript paths.** `walkFiles` joined with a literal `/`, returning mixed separators on Windows. Reads still worked, so it stayed invisible — but the string never equalled a `path.join`-built one, and `listTranscriptFiles` keys its mtime map by exactly that string, so one transcript could be tracked under two spellings. Transcript files are now matched on basename rather than an `endsWith("/name.jsonl")` suffix, because that suffix test only worked *while* the paths were wrongly built; fixing one without the other would have silently stopped recognising every transcript on Windows.

**Tests narrowed to what each platform can express.** Twelve owner-only mode assertions route through `expectPosixMode`, which checks the bits on POSIX and skips where the platform cannot express them. This narrows the tests to what is true; it does not make the property hold, and the helper's doc-comment says so at the point anyone will read it. Five `stop`/`stopAndWait` cases that attribute a bare spawned listener via its environment skip on Windows with the reason recorded, and still run on POSIX.

Two fixtures were wrong rather than the code: `resolveInstanceHome` correctly returns native separators, and `os.homedir()` follows `%USERPROFILE%` on Windows, so the instance-aware test was reading the operator's real `~/.jinn` instead of its fixture. And the migration-bundle fixture inherited the developer's `core.autocrlf`, which put git's "LF will be replaced by CRLF" advisory into a `stderr === ""` assertion — a throwaway repo should pin its own line-ending config.

**The suite was under-reporting.** Windows has no `fork()`, so each vitest pool worker is a full process spawn re-resolving the module graph over NTFS with realtime AV scanning. Suites whose `beforeAll` awaits an import of `api.ts` — 350KB pulling in most of the gateway — routinely passed the 10s hook timeout. Vitest reports a timed-out hook by marking that file's tests **skipped, not failed**, so roughly twenty suites (~180 tests, including the work-item and privileged-read guards) were dropping out of local Windows runs while the run still reported green. The config already grants this headroom on CI for the same underlying reason.

**On Windows CI:** agreed it is the highest-leverage thing here, and I have the one blocker fixed (the migration-bundle fixture above was the last non-flake Windows failure). I have not opened that PR yet — it needs to stack on this one, since `pnpm build` cannot pass on a Windows runner without it.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,38 @@
+# Line endings are load-bearing in this repository.
+#
+# Shipped migration payloads under packages/jinn/template/migrations are
+# content-addressed: each manifest stores a sha256 of the payload bytes, and the
+# migration snapshot compares payloads byte-for-byte against the user's instance
+# files. A contributor with core.autocrlf=true (the Git for Windows default) got
+# CRLF working-tree copies, so every one of those hashes mismatched and the
+# migration service failed to validate its own shipped bundle.
+#
+# Normalize to LF in the repository and in the working tree on every platform.
+* text=auto eol=lf
+
+# Windows script hosts require CRLF.
+*.bat text eol=crlf
+*.cmd text eol=crlf
+*.ps1 text eol=crlf
+
+# Never touch binary payloads.
+*.png  binary
+*.jpg  binary
+*.jpeg binary
+*.gif  binary
+*.webp binary
+*.ico  binary
+*.svg  text eol=lf
+*.mp4  binary
+*.webm binary
+*.mov  binary
+*.woff  binary
+*.woff2 binary
+*.ttf  binary
+*.otf  binary
+*.node binary
+*.wasm binary
+*.db   binary
+*.sqlite binary
+*.gz   binary
+*.zip  binary
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "test": "turbo run test --concurrency=1",
     "coverage": "turbo run coverage",
     "test:e2e": "playwright test",
-    "clean": "rm -rf .turbo node_modules/.cache && turbo clean",
+    "clean": "node -e \"const fs=require('fs');for(const d of ['.turbo','node_modules/.cache'])fs.rmSync(d,{recursive:true,force:true})\" && turbo clean",
     "jinn": "node packages/jinn/dist/bin/jinn.js",
     "setup": "pnpm build && node packages/jinn/dist/bin/jinn.js setup",
     "setup:force": "pnpm build && node packages/jinn/dist/bin/jinn.js setup --force",
--- a/packages/jinn/package.json
+++ b/packages/jinn/package.json
@@ -20,7 +20,7 @@
   ],
   "scripts": {
     "postinstall": "node scripts/fix-node-pty-permissions.mjs",
-    "build": "rm -rf dist/bin dist/src && tsc -p tsconfig.build.json && mkdir -p dist/src/talk && (cp src/talk/*.md src/talk/*.py dist/src/talk/ 2>/dev/null || true)",
+    "build": "node scripts/build.mjs",
     "dev": "tsc && (tsc --watch &) && node --watch-path=dist dist/bin/jinn.js start",
     "typecheck": "tsc --noEmit",
     "test": "pnpm build && vitest run",
@@ -28,7 +28,7 @@
     "test:watch": "vitest",
     "migration:generate": "node scripts/instance-migration-bundle.mjs generate",
     "migration:check": "node scripts/instance-migration-bundle.mjs check",
-    "clean": "rm -rf dist"
+    "clean": "node -e \"require('fs').rmSync('dist',{recursive:true,force:true})\""
   },
   "dependencies": {
     "@slack/bolt": "^4.7.3",
--- a/packages/jinn/scripts/build.mjs
+++ b/packages/jinn/scripts/build.mjs
@@ -1,0 +1,55 @@
+#!/usr/bin/env node
+/**
+ * Cross-platform build for jinn-cli.
+ *
+ * This was a shell one-liner (`rm -rf … && tsc && mkdir -p … && cp …`). npm and
+ * pnpm run package scripts through the platform shell, which on Windows is cmd —
+ * where none of those commands exist, so `pnpm build` failed with "The syntax of
+ * the command is incorrect" and the repo could not be built there at all. Node's
+ * own fs APIs do the same work on every platform.
+ */
+import fs from "node:fs";
+import path from "node:path";
+import { spawnSync } from "node:child_process";
+import { createRequire } from "node:module";
+import { fileURLToPath } from "node:url";
+
+const packageRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "..");
+const dist = path.join(packageRoot, "dist");
+const require = createRequire(import.meta.url);
+
+/** Run TypeScript's own entry with the current Node. Resolving the JS entry
+ *  instead of the `tsc` shim keeps this shell-free, which avoids both the
+ *  Windows .CMD resolution problem and shell argument concatenation. */
+function runTsc(args) {
+  const tsc = require.resolve("typescript/bin/tsc");
+  const result = spawnSync(process.execPath, [tsc, ...args], { cwd: packageRoot, stdio: "inherit" });
+  if (result.error) throw result.error;
+  if (result.status !== 0) process.exit(result.status ?? 1);
+}
+
+// 1. Clear only the compiled output. dist/web is produced separately by
+//    scripts/sync-web-dist.mjs and must survive a package rebuild.
+for (const dir of ["bin", "src"]) {
+  fs.rmSync(path.join(dist, dir), { recursive: true, force: true });
+}
+
+// 2. Compile.
+runTsc(["-p", "tsconfig.build.json"]);
+
+// 3. Copy the Talk assets tsc does not emit (Markdown + Python live beside the
+//    TypeScript). Missing files are not an error: they are optional extras.
+const talkSource = path.join(packageRoot, "src", "talk");
+const talkTarget = path.join(dist, "src", "talk");
+fs.mkdirSync(talkTarget, { recursive: true });
+let copied = 0;
+try {
+  for (const entry of fs.readdirSync(talkSource)) {
+    if (!/\.(md|py)$/i.test(entry)) continue;
+    fs.copyFileSync(path.join(talkSource, entry), path.join(talkTarget, entry));
+    copied += 1;
+  }
+} catch (error) {
+  if (error.code !== "ENOENT") throw error;
+}
+console.log(`build: compiled to dist/, copied ${copied} talk asset(s)`);
--- a/packages/jinn/src/cli/__tests__/instance-migration-bundle.test.ts
+++ b/packages/jinn/src/cli/__tests__/instance-migration-bundle.test.ts
@@ -24,6 +24,12 @@ function fixture(): string {
   git(root, "init", "-q")
   git(root, "config", "user.name", "Test Operator")
   git(root, "config", "user.email", "test@example.invalid")
+  // A throwaway fixture must not inherit the developer's line-ending config.
+  // With core.autocrlf=true (the Git for Windows default) git prints "LF will be
+  // replaced by CRLF" advisories to stderr, and one case below asserts stderr is
+  // empty — so the suite failed on Windows for a property it never meant to test.
+  git(root, "config", "core.autocrlf", "false")
+  git(root, "config", "core.eol", "lf")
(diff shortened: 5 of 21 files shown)


hristo2612:
Reviewed in depth. **Recommendation: SPLIT.** This bundles three concerns with three different readiness levels, and holding the safe one hostage to the risky one helps nobody. Rebase is clean (your branch's merge-base is current main, so v0.28.6's `6c95c2d4` is already an ancestor).

Test run on macOS, `pnpm --filter jinn-cli test`: **3493 passed, 1 failed** — `owner-only.test.ts > pathIsOwnerOnly > fails closed for a path that does not exist` is a genuine PR-caused failure (see ③). The `activity/performance.test.ts` timeout is a pre-existing flake; it passes standalone.

---

## ① "Make the repo buildable" — **merge-ready**, land this now

`pnpm build`/`clean` died on Windows because npm runs scripts through `cmd.exe`, which has no `rm -rf`/`mkdir -p`/`cp`/`2>/dev/null`. Verified `scripts/build.mjs` produces output identical to the old shell one-liner, typecheck clean. `.gitattributes` verified harmless — `git add --renormalize .` produces zero changes and no CRLF blobs exist in the index. The `grok-interactive.ts` `path.join` change is strictly more correct on POSIX too and fixes a real double-tracking bug in the mtime map.

**One nit:** `packages/jinn/src/shared/test-support/posix-mode.ts` imports `vitest` but isn't excluded by `tsconfig.build.json` (`exclude` only covers `**/__tests__/**` and `**/*.test.ts`). It emits to `dist/` and ships in the npm tarball. Add `**/test-support/**`.

## ② "Port detection correct" — **after changes**, and less risky than it looks

To be clear about what this does *not* do, since the title made me nervous: it doesn't touch `assertPidBelongsToThisInstance`, `readProcessJinnHome`, or the `gateway.json` fallback. No conflict with `2dabcb23` or `6c95c2d4`.

It **closes a destructive bug**: `lookupPidOnPort` took the PID from the first `findstr LISTENING` row and ignored the bind address entirely — the same defect `2dabcb23` fixed for `lsof` but never for `netstat` — and that PID is fed to `process.kill` on the `--take-port` and force-kill paths. I exercised your exported parsers against synthetic `netstat -ano` tables including a German `ABHÖREN` row: 8/8 correct (picks the right loopback listener, ignores ESTABLISHED/TIME_WAIT and `pid 0`, skips 4-column UDP rows, handles `[::1]:7777`, normalises `0.0.0.0`/`[::]`, genuinely state-word-independent).

Required before merge:

1. **Ship tests.** `parseNetstatAddress` and `selectNetstatPortOwnerPid` are exported with no test file anywhere — the riskiest code in the PR is the only risky code you didn't test. They're pure functions, so they run on Linux CI. The synthetic-table cases above are exactly what's missing.
2. **Restore `unknown` on a total parse miss.** You changed a non-empty-but-unparseable result from `{status:"unknown"}` (→ safe refusal) to `{status:"none"}` (= "port is free"). For a genuinely non-overlapping listener that's correct. But if some Windows SKU or locale renders a shape the parser can't read *at all*, we now report the port free and walk into an EADDRINUSE crash instead of a clean error. If **no** line parsed into a valid address, return `unknown`.
3. **`execFileSync("netstat", ["-ano"])` needs `maxBuffer`, `timeout`, `windowsHide`.** The old command piped through `findstr` so Node buffered a few bytes; now it buffers the whole connection table against Node's 1 MB default (~17k rows). `waitForPortFree` calls this every 200 ms for up to 10 s. Your own `owner-only.ts` passes `windowsHide: true` to `icacls` — inconsistent, and this can flash a console window.
4. **`c108d846`'s commit message is stale** — it still claims it "checks our own `gateway.json` first", which is `6c95c2d4`'s work absorbed by your rebase, not in this diff. A reviewer skimming messages would think this PR touches ownership attribution.

## ③ "Instance files owner-only" — **hold, split into its own PR**

This is real ACL hardening, not a loosened check — credit where due, and it does subsume **issue #99** (the `mode === 0o600` equality that can't be satisfied on NTFS). But it has four problems, and you flagged this commit yourself as "the one to argue about. You were right.

1. **`pathIsOwnerOnly` doesn't fail closed on POSIX — it throws.** `fs.statSync` raises ENOENT on a missing path; the module's own doc-comment says "Everything here fails closed" and its own test asserts `false`. This is the failing test above, and it's a live crash path: `server.ts` calls `pathIsOwnerOnly(JINN_HOME)` unguarded at gateway boot **on all platforms**. Fix: `fs.statSync(target, { throwIfNoEntry: false })`.
2. **Hardcoded English principal names — the exact bug class you fixed in netstat.** `ALWAYS_PRIVILEGED` uses `NT AUTHORITY\SYSTEM` / `BUILTIN\ADMINISTRATORS`, and `/grant:r` passes those literals. On German Windows `icacls` prints `VORDEFINIERT\Administratoren`, so `pathIsOwnerOnly` returns `false` for every path (permanent boot warning) **and #99 stays broken on localized Windows**. Use SIDs: `*S-1-5-18`, `*S-1-5-32-544`, `*S-1-3-0`, `*S-1-3-4`. Your own test already grants by SID with a comment noting names are localized — the production code just didn't follow through.
3. **Undisclosed POSIX-wide behaviour change.** `enforceOwnerOnlyDirectory` chmods `0o700` on POSIX and `jinn setup` now calls it. The default `~/.jinn` is `0755` (mkdir under umask 022), so on merge **every existing macOS/Linux user gets a security warning on every gateway boot** until they re-run setup. Defensible security-wise, but it's a user-visible change shipped under a PR titled "fix(windows)". Needs a deliberate decision, a changelog entry, and probably a one-shot self-heal or a quieter first-run message.
4. Minor: `currentWindowsPrincipal()` reads `%USERNAME%`/`%USERDOMAIN%` rather than the process token — same trust level as the parent shell, so not an escalation, but a SID lookup is more robust.

Your test-narrowing (12 `expectPosixMode` conversions + 6 skips) is honest and well-documented — the helper's doc-comment says plainly that the property doesn't hold on Windows, and no POSIX coverage is lost. The `vitest.config.ts` timeout bump is correctly gated to win32.

---

## The bigger blocker

`gh pr checks 97` reports **no checks at all**, because `.github/workflows/ci.yml` is `ubuntu-latest` on all three jobs. There is no Windows CI, so every claim in this PR is unverifiable today and will silently rot — your `isWindows` timeout bump and the ACL paths have no guard whatsoever. Your PR #96 demonstrates the failure mode: its Windows tests are `skipIf`'d against the real platform and therefore never execute.

I'd like a `windows-latest` matrix leg to land alongside or before this work. Would you be up for that as a separate PR? It's the highest-leverage thing in this batch, and it's the reason ① can merge quickly while ②/③ can't.

Genuinely good work here — the netstat bind-address bug in particular is one I'd have missed.

e1010101:
Split as you asked. This PR is now ① only.

- **② port detection → #100.** All four required changes: netstat parser tests (pure functions, so they run on Linux CI, including your German `ABHÖREN` case and the pid-0/TIME_WAIT/UDP rows), `unknown` restored on a total parse miss, `maxBuffer`/`timeout`/`windowsHide`, and the stale commit message removed.
- **③ owner-only → #101.** Stacked on this one because it needs `expectPosixMode`. All four points addressed — `throwIfNoEntry: false`, SIDs on both the read and the grant side, and the POSIX `chmod 0700` change now disclosed as a `BEHAVIOUR CHANGE ON POSIX` in the commit message.

Your `tsconfig.build.json` nit is fixed here: `test-support` was emitting into `dist/` and shipping in the tarball. `**/test-support/**` is excluded and I confirmed `dist/src/shared/test-support` is gone after a build.

One thing your review made me chase that turned out to matter more than expected. While rebasing onto v0.28.6 I upgraded my own instance and the CLI and gateway computed **different migration keys for the same upgrade** — because my local `0.28.5`/`0.28.6` bundles were CRLF while the published package ships LF. Same defect this PR fixes, hitting for real rather than in a test. It also exposed a limitation worth stating: `.gitattributes` fixes fresh clones, but files arriving later via `git fetch` onto a branch that does not carry it still land CRLF, so an existing checkout needs one renormalisation pass. That is now in the description.

On **Windows CI**: agreed it is the highest-leverage item. I have the last non-flake Windows failure fixed here — the `instance-migration-bundle` fixture was inheriting the developer's `core.autocrlf`, putting git's advisory into a `stderr === ""` assertion. I have not opened that PR yet; it has to stack on this one, since `pnpm build` cannot pass on a Windows runner without it. Happy to do it next if you want it before ②/③.

e1010101:
Two follow-ups.

**One more Windows-only test bug, fixed here.** `comms-v2.test.tsx`'s `strippedLines` splits source on `'\n'`, which leaves a trailing `\r` on a CRLF checkout — and `.` never matches `\r`, so the comment-stripping `/(^|[^:])\/\/.*$/` cannot reach the end of the line and strips nothing. Every commented em dash in the chat and ui component sources then reports as rendered copy: 12 offenders in `chat-sidebar.tsx` alone, all ordinary code comments. `.gitattributes` prevents the CRLF checkout that triggers it, but the check should not depend on that, so it now splits on `/\r?\n/`.

**Integration check across all six PRs.** I merged #94, #95, #96, #97, #100 and #101 into one local branch to see how they combine:

- They merge cleanly except for one three-line conflict, flagged below.
- `pnpm typecheck` is clean for both packages.
- Full `packages/jinn` suite on Windows 11 / Node 24: **3520 passed, 4 failed, 10 skipped**, plus load-related flakiness described below.

The 4 failures are **pre-existing on `main`** — I confirmed by running the same two files on `main`, with an identical result — and none of these PRs claim them:

- `engine-limits.test.ts`, three codex-rollout tests. `writeCodexAppServerStub` writes a `#!/usr/bin/env node` file and `chmod 0o755`; Windows has no shebang, so `spawn(bin, ["app-server", "--stdio"])` never starts it and `expect(fs.existsSync(startedFile)).toBe(true)` fails in all three.
- `pty-stream.test.ts`, `bounds stalled pre-ready control output`. Times out at 30s; it pushes ~2,100 control sequences plus 512KB through xterm-headless.

Worth knowing separately: the same shebang detail points at a **production** gap, not just a test one. `engine-limits.ts:355` uses a bare `spawn(bin, ...)` and `:199` a bare `execFile(bin, ...)`. An npm-installed `codex` on Windows is `codex.cmd`, which Node refuses to spawn without a shell — the same constraint `cli/skills.ts:145` and `setup.ts:673` already work around. So codex usage limits are likely unavailable on Windows regardless of these tests. I have not touched it; happy to open a separate PR if you want it.

There is also genuine flakiness on Windows under full-suite parallelism, unrelated to any of these branches: `SqliteError: database is locked` between suites, `EPERM` removing the vitest temp root during teardown, and occasional 60s hook timeouts. A different subset fails per run. That is what a Windows CI job would have to contend with beyond the 4 deterministic failures, which is why I have not proposed one yet.

**Merge-order note:** #95 and this PR both touch `snapshot.test.ts:135-138`. This PR converts the mode assertion to `expectPosixMode`; #95 replaces the line above it, which asserted a materialized symlink, with the opposite assertion. Whichever lands second takes a three-line conflict. The resolution is to keep both changes:

```ts
expect(fs.existsSync(path.join(first.path, "AGENTS.md"))).toBe(false)
expectPosixMode(fs.statSync(path.join(first.path, "CLAUDE.md")), 0o640)
```

Issue #99, opened 2026-07-26
[Windows] `jinn pair` always fails with "invalid challenge" — POSIX 0600 check on proof file


kaiengel:
# [Windows] `jinn pair` always fails with "invalid challenge" — POSIX 0600 check on proof file

## Environment
- jinn-cli 0.28.2 – 0.28.6, Windows 11, NTFS

## Symptom
CLI pairing never succeeds on Windows. The gateway rejects every attempt with
"invalid challenge", regardless of token correctness.

## Root cause
The gateway validates that the pairing proof file has POSIX mode exactly `0600`
(gateway/pairing-challenge.*). On Windows, Node's `fs.chmod` can only toggle the
read-only bit: `fs.statSync(...).mode & 0o777` reports `0666` (or `0444` after
chmod 0400), never `0600`. The equality check therefore fails on every Windows
system by construction.

## Repro
1. Windows, any jinn instance with `gateway.authRequired: true`.
2. Run `jinn pair` and complete the flow.
3. Gateway logs/returns "invalid challenge" every time.

## Suggested fix
Skip the strict mode check on Windows (`process.platform !== "win32" && mode !== 0o600`)
or replace it with an ACL/ownership check there. File modes are a POSIX concept; on NTFS
the 0600 guarantee cannot be expressed via `chmod`.

## Impact
Low for the happy path (foreground start auto-pairs via bootstrap grant in the opened
browser), but it blocks headless/remote pairing setups on Windows entirely.

hristo2612:
Confirmed — the claim is correct and the failure is by construction, not environmental. Thanks for the precise diagnosis.

`packages/jinn/src/gateway/pairing-challenge.ts:50`:

```ts
if ((proofStat.mode & 0o777) !== 0o600) return false;
```

Strict equality after masking, no win32 branch anywhere in the gateway's pairing code. Since `fs.chmod` on Windows only toggles the read-only attribute, libuv derives `st_mode` from `FILE_ATTRIBUTE_READONLY` alone — writable stats as `0o666`, read-only as `0o444`. `0o600` is unreachable, so `consumePairingChallenge` always returns false and the route answers "invalid challenge" every time.

[...]

PR #82, opened 2026-07-21
fix(cli): restore node-pty Darwin helper modes after install

juyoungk23:
This fixes an issue where every time i restarted the gateway, my Claude engine wouldn't work. I'd have to reconnect the Claude engine. Codex was working fine after restarts for me because i am using a headless codex. this should ensure that Claude will continue working even after gateway restarts. 

Fixes macOS npm installs that leave node-pty spawn helpers non-executable. The postinstall resolves node-pty from the installed CLI context, chmods Darwin helpers, and fails explicitly on resolution/prebuild/helper errors.

Validation: typecheck + build; packed fresh hoisted install observed 0644 before, 0755 after; node-pty spawn(`/bin/echo`, [`PTY_OK`]) passed.

PR #101, opened 2026-07-26
fix(security): make instance files owner-only, by SID rather than by name

e1010101:
The instance holds the gateway token, connector secrets and every session transcript, protected with mode `0o600`. Windows has no POSIX mode bits — `chmod` there only toggles the read-only attribute — so that call is a no-op and the home keeps whatever ACL it inherits from the profile. A group granted access on the profile reads the entire instance. This also subsumes **#99**, the `mode === 0o600` equality that no NTFS file can satisfy.

> Split out of #97 as ③, per review — this is the one to argue about. **Stacked on #97**, because it needs the `expectPosixMode` helper introduced there; review or merge that first. All four of your points are addressed below, including the POSIX behaviour change you were right to call undisclosed.

Found on a real instance: every file under `~/.jinn` readable by a sandbox account group, including `gateway.json`, `secrets/` and `sessions/registry.db`.

---

**1. Fails closed instead of throwing.** `fs.statSync` raises ENOENT, and `server.ts` calls `pathIsOwnerOnly(JINN_HOME)` unguarded at boot **on every platform** — a live crash path that contradicted both the module's stated contract and its own test. Now `fs.statSync(target, { throwIfNoEntry: false })` with a null check, and the test asserts it.

**2. SIDs throughout, not English names.** You were right that this was the same bug class I had just fixed in netstat, in my own code. Fixed on both sides:

- **Reading** no longer parses `icacls`'s human output at all. It reads the descriptor as SDDL via `icacls /save`, whose trustees are well-known two-letter aliases (`SY`, `BA`) and raw SIDs — locale-independent by construction. `parseIcaclsPrincipals` is gone, replaced by `parseSddlTrustees`.
- **Granting** passes the `*S-…` form: `*S-1-5-18`, `*S-1-5-32-544`, plus the caller's own SID. The literals would have had to be the localised ones on a non-English Windows, where `BUILTIN\Administrators` does not resolve and the whole call fails.
- Identity comes from the SID via `whoami /user`, not `%USERNAME%`/`%USERDOMAIN%` — addressing your fourth point too, since a name still has to be resolved and resolution is exactly what localisation breaks.

One thing that fell out of testing this: `whoami` and `icacls` are now resolved from `%SystemRoot%\System32` rather than `PATH`. MSYS, Cygwin and Git Bash ship their own `whoami` ahead of Windows' one, and a bare call gets the Unix binary, fails to parse, and silently degrades the check to "cannot tell" — a permanent boot warning plus a hardening call that quietly does nothing. I hit exactly that while verifying.

Verified end-to-end on real paths: the SID resolves, a temp directory inheriting a shared `%TEMP%` ACL reads as *not* owner-only, `enforceOwnerOnlyDirectory` makes it owner-only, a newly created child inherits that, and a missing path returns false without throwing.

**3. The POSIX behaviour change, now disclosed.** `enforceOwnerOnlyDirectory` chmods `0o700` on POSIX and `jinn setup` calls it. The default `~/.jinn` is `0755` under a 022 umask, so on merge **every existing macOS and Linux instance logs the boot warning** until setup is re-run or the directory is tightened by hand. It is the correct posture for a directory holding a gateway token, but it is user-visible and shipped under a Windows-titled PR — so it is now stated in the commit message as a `BEHAVIOUR CHANGE ON POSIX` and wants a changelog entry. Boot only warns and never repairs, deliberately: an operator may have granted that access on purpose, and rewriting ACLs under a running install could break it.

Per your comment, the one-shot self-heal is now in: on POSIX only, boot tightens an over-permissive `JINN_HOME` to `0700` once, writes a `.owner-only-healed` marker so it never fights an operator who deliberately widens it again, and logs what it did. Windows still only warns — rewriting ACLs under a running install is not something to do unasked, and the audit that started this found a group grant that may well be intentional. The user-visible `0755 → 0700` change is recorded in the changelog under Security.

**Also here:** the pairing proof required mode `0o600` unconditionally, which no Windows file can report, so local pairing was impossible there (#99). It now reads the ACL on Windows and the mode/uid triple on POSIX. Both pairing fixtures establish an owner-only home explicitly, since `mkdtemp` gives that for free on POSIX but not on Windows, where a temp directory inherits whatever `%TEMP%` grants — often several groups with Modify.

Tests cover the SDDL parser directly (it decides whether a path counts as shared): trustee extraction, a group that must be rejected, Deny ACEs ignored, malformed ACEs skipped, and empty output failing closed. Those are platform-independent and run on CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [0.28.8]
+
+### 🔒 Security
+- **The instance directory is restricted to you, and says so when it is not.** `~/.jinn` holds the gateway token, connector secrets and every session transcript, and was protected with a `0600` file mode. Windows has no such mode — `chmod` there only toggles the read-only flag — so the protection was a no-op and the directory kept whatever access it inherited from the user profile; on one machine that meant every file readable by an unrelated account group. Permissions are now expressed in each platform's own model, identified by security identifier rather than by name so that a non-English Windows is handled correctly, and `jinn setup` restricts the directory once so everything written later inherits it. Local device pairing also worked again on Windows, where it had required a file mode no Windows file can report.
+
+  **Existing macOS and Linux instances are affected.** A directory created before this release is group- and world-readable (`0755`), so the gateway now tightens it to `0700` once, at startup, and logs that it did. Widen it again deliberately and the gateway reports the exposure on each start instead of overriding you. Windows is never repaired automatically: doing so means dropping inherited access an operator may have granted on purpose, such as a sandboxed engine account that needs to read configuration.
+
 ## [0.28.7] - 2026-07-27
 
 ### 🐛 Fixes
--- a/packages/jinn/src/cli/setup.ts
+++ b/packages/jinn/src/cli/setup.ts
@@ -5,6 +5,7 @@ import readline from "node:readline";
 import { execSync, spawn } from "node:child_process";
 import yaml from "js-yaml";
 import { isInstalled, resolveBin } from "../shared/resolve-bin.js";
+import { enforceOwnerOnlyDirectory } from "../shared/owner-only.js";
 import {
   JINN_HOME,
   CONFIG_PATH,
@@ -495,6 +496,17 @@ export async function runSetup(opts?: { force?: boolean }): Promise<void> {
   const created: string[] = [];
 
   if (ensureDir(JINN_HOME)) created.push(JINN_HOME);
+
+  // Restrict the instance to this user before anything sensitive is written into
+  // it. Done once on the directory, with inheritance, so every file created later
+  // is covered: on Windows there are no mode bits to set per file, and a home
+  // that inherits a broad ACE from the profile would expose the gateway token,
+  // connector secrets and the whole session database to those principals.
+  if (enforceOwnerOnlyDirectory(JINN_HOME)) {
+    ok(`Restricted ${JINN_HOME} to ${process.platform === "win32" ? "your account" : "owner-only (0700)"}`);
+  } else {
+    warn(`Could not restrict permissions on ${JINN_HOME}; check who can read it before adding credentials`);
+  }
 
   // Copy or create config files.
   // DEFAULT_CONFIG (above) is the canonical default. `template/config.yaml` is an
--- a/packages/jinn/src/gateway/__tests__/auth-ux-api.test.ts
+++ b/packages/jinn/src/gateway/__tests__/auth-ux-api.test.ts
@@ -6,6 +6,7 @@ import { Readable } from "node:stream";
 import type { IncomingMessage, ServerResponse } from "node:http";
 import { handleApiRequest, type ApiContext } from "../api.js";
 import { requestPairingCode } from "../../cli/pair.js";
+import { enforceOwnerOnlyDirectory } from "../../shared/owner-only.js";
 import {
   createAuthSession,
   issueLocalBootstrapGrant,
@@ -69,7 +70,16 @@ function makeRes() {
   };
 }
 
-function ctx(jinnHome = fs.mkdtempSync(path.join(os.tmpdir(), "jinn-auth-api-"))): ApiContext {
+/** The loopback pairing proof requires a home only this account can reach.
+ *  mkdtemp gives that on POSIX (0700) but not on Windows, where a temp
+ *  directory inherits whatever %TEMP% grants. Establish it either way. */
+function ownerOnlyTempHome(prefix: string): string {
+  const home = fs.mkdtempSync(path.join(os.tmpdir(), prefix));
+  enforceOwnerOnlyDirectory(home);
+  return home;
+}
+
+function ctx(jinnHome = ownerOnlyTempHome("jinn-auth-api-")): ApiContext {
   return {
     gatewayAuthToken: "gateway-token",
     jinnHome,
--- a/packages/jinn/src/gateway/__tests__/pairing-challenge.test.ts
+++ b/packages/jinn/src/gateway/__tests__/pairing-challenge.test.ts
@@ -1,7 +1,10 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
+import { execFileSync } from "node:child_process";
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
+import { expectPosixMode } from "../../shared/test-support/posix-mode.js";
+import { enforceOwnerOnlyDirectory } from "../../shared/owner-only.js";
 import {
   consumePairingChallenge,
   issuePairingChallenge,
@@ -10,7 +13,14 @@ import {
 } from "../pairing-challenge.js";
 
 function tempHome(): string {
-  return fs.mkdtempSync(path.join(os.tmpdir(), "jinn-pair-challenge-"));
+  const home = fs.mkdtempSync(path.join(os.tmpdir(), "jinn-pair-challenge-"));
+  // The proof is only meaningful in a home nobody else can reach, and these
+  // cases assume that premise. mkdtemp gives it for free on POSIX (0700), but
+  // not on Windows, where a temp directory inherits whatever %TEMP% grants —
+  // often several groups with Modify. Establish it explicitly so the fixture
+  // matches the precondition on both platforms.
+  enforceOwnerOnlyDirectory(home);
+  return home;
 }
 
 function issue(home: string, store: PairingChallengeStore, now = 1_000) {
@@ -36,7 +46,7 @@ describe("pairing filesystem challenges", () => {
     fs.chmodSync(challenge.path, 0o600);
 
     expect(fs.statSync(challenge.path).uid).toBe(fs.statSync(home).uid);
-    expect(fs.statSync(challenge.path).mode & 0o777).toBe(0o600);
+    expectPosixMode(fs.statSync(challenge.path), 0o600);
     expect(consumePairingChallenge(home, challenge.challengeId, store, 1_001)).toBe(true);
     expect(fs.existsSync(challenge.path)).toBe(false);
     expect(consumePairingChallenge(home, challenge.challengeId, store, 1_002)).toBe(false);
@@ -70,9 +80,19 @@ describe("pairing filesystem challenges", () => {
     const store: PairingChallengeStore = new Map();
     const broad = issue(home, store);
     fs.writeFileSync(broad.path, broad.nonce, { mode: 0o644 });
-    fs.chmodSync(broad.path, 0o644);
+    if (process.platform === "win32") {
+      // chmod cannot widen anything on Windows. Grant the Everyone SID instead -
+      // by SID, because the name is localized - which is what "readable by more
+      // than the owner" actually means on this platform.
+      execFileSync("icacls", [broad.path, "/grant", "*S-1-1-0:(R)"], { windowsHide: true });
+    } else {
+      fs.chmodSync(broad.path, 0o644);
+    }
     expect(consumePairingChallenge(home, broad.challengeId, store, 1_001)).toBe(false);
 
+    // A forged uid is a POSIX-only notion: the Windows check reads the ACL, and
+    // ownership cannot be reassigned to another account from inside a test.
+    if (process.platform !== "win32") {
     const foreign = issuePairingChallenge(
       home,
       store,
@@ -91,6 +111,7 @@ describe("pairing filesystem challenges", () => {
     }) as typeof fs.lstatSync);
     expect(consumePairingChallenge(home, foreign.challengeId, store, 2_001)).toBe(false);
     vi.restoreAllMocks();
+    }
 
     const symlink = issuePairingChallenge(
       home,
--- a/packages/jinn/src/gateway/__tests__/pairing-challenge.test.ts
+++ b/packages/jinn/src/gateway/__tests__/pairing-challenge.test.ts
@@ -4,7 +4,7 @@ import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 import { expectPosixMode } from "../../shared/test-support/posix-mode.js";
-import { enforceOwnerOnlyDirectory } from "../../shared/owner-only.js";
+import { currentWindowsSid, enforceOwnerOnlyDirectory, parseSddlTrustees, pathIsOwnerOnly } from "../../shared/owner-only.js";
 import {
   consumePairingChallenge,
   issuePairingChallenge,
@@ -20,6 +20,34 @@ function icacls(): string {
   return path.join(root, "System32", "icacls.exe");
 }
 
+/** What the owner-only check actually sees, for failure messages.
+ *
+ *  A bare "expected false to be true" from these cases is close to undiagnosable
+ *  on a machine you cannot log into: it could be identity resolution, the ACL
+ *  edit, or inheritance onto the proof file. Naming the state costs one icacls
+ *  call on the failure path only. */
+function ownerOnlyDiagnosis(target: string): string {
+  if (process.platform !== "win32") {
+    const stat = fs.statSync(target, { throwIfNoEntry: false });
+    return stat ? `mode=0o${(stat.mode & 0o777).toString(8)} uid=${stat.uid}` : "missing";
+  }
+  const dump = path.join(os.tmpdir(), `pair-diag-${process.pid}-${Math.abs(hashPath(target))}.sddl`);
+  try {
+    execFileSync(icacls(), [target, "/save", dump, "/Q"], { windowsHide: true, stdio: ["ignore", "pipe", "ignore"] });
+    return `sid=${currentWindowsSid() ?? "UNRESOLVED"} trustees=[${parseSddlTrustees(fs.readFileSync(dump, "utf16le")).join(", ")}]`;
+  } catch (error) {
+    return `sid=${currentWindowsSid() ?? "UNRESOLVED"} icacls-save-failed: ${(error as Error).message.slice(0, 120)}`;
+  } finally {
+    try { fs.unlinkSync(dump); } catch { /* nothing written */ }
+  }
+}
+
+function hashPath(value: string): number {
+  let h = 0;
+  for (let i = 0; i < value.length; i += 1) h = (h * 31 + value.charCodeAt(i)) | 0;
+  return h;
+}
+
 function tempHome(): string {
   const home = fs.mkdtempSync(path.join(os.tmpdir(), "jinn-pair-challenge-"));
   // The proof is only meaningful in a home nobody else can reach, and these
@@ -27,7 +55,11 @@ function tempHome(): string {
   // not on Windows, where a temp directory inherits whatever %TEMP% grants —
   // often several groups with Modify. Establish it explicitly so the fixture
   // matches the precondition on both platforms.
-  enforceOwnerOnlyDirectory(home);
+  //
+  // Asserted rather than assumed: if this fails, every acceptance case below
+  // fails with "expected false to be true" and the real cause is invisible.
+  const applied = enforceOwnerOnlyDirectory(home);
+  expect(applied, `could not establish an owner-only home at ${home} — ${ownerOnlyDiagnosis(home)}`).toBe(true);
   return home;
 }
 
@@ -55,6 +87,10 @@ describe("pairing filesystem challenges", () => {
 
     expect(fs.statSync(challenge.path).uid).toBe(fs.statSync(home).uid);
     expectPosixMode(fs.statSync(challenge.path), 0o600);
+    // The proof inherits the home's restriction rather than setting its own on
+    // Windows, so check that separately: it is the half a mode-only assertion
+    // cannot see, and the half that fails if inheritance did not apply.
+    expect(pathIsOwnerOnly(challenge.path), `proof file is not owner-only — ${ownerOnlyDiagnosis(challenge.path)}`).toBe(true);
     expect(consumePairingChallenge(home, challenge.challengeId, store, 1_001)).toBe(true);
     expect(fs.existsSync(challenge.path)).toBe(false);
     expect(consumePairingChallenge(home, challenge.challengeId, store, 1_002)).toBe(false);
--- a/packages/jinn/src/gateway/__tests__/pairing-challenge.test.ts
+++ b/packages/jinn/src/gateway/__tests__/pairing-challenge.test.ts
@@ -12,6 +12,14 @@ import {
   type PairingChallengeStore,
 } from "../pairing-challenge.js";
 
+/** Resolve icacls from System32, never PATH. MSYS/Cygwin/Git Bash ship their own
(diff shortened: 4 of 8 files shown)


e1010101:
Took your advice on point 3 as written — deliberate decision, changelog entry, and a one-shot self-heal.

**Self-heal, POSIX only, once.** `chmod 700` on a directory we own is cheap, reversible, and exactly what `jinn setup` does, so boot now does it rather than warning every existing macOS/Linux operator forever. A marker file records the attempt, so someone who re-widens the directory deliberately gets the warning instead of being silently overridden.

Windows stays warn-only, and the asymmetry is the whole point: repairing there means `/inheritance:r`, which drops inherited ACEs an operator may have granted on purpose — a sandboxed engine account that needs to read config — and doing that under a running install could break it. On POSIX there is no equivalent inherited-grant mechanism to destroy, so the risk profiles genuinely differ.

**Changelog entry** added under `[Unreleased]`, with the POSIX-visible change called out in its own paragraph rather than buried: existing instances are `0755`, the gateway tightens once and says so, and Windows is never repaired automatically.

Points 1, 2 and 4 were already addressed in the pushed branch — `throwIfNoEntry: false`, SIDs on both the read (`icacls /save` SDDL) and grant (`*S-…`) sides, and identity from the SID rather than `%USERNAME%`. Worth flagging one extra fix that fell out of verifying it: `whoami` and `icacls` are resolved from `%SystemRoot%\System32` rather than `PATH`, because MSYS/Cygwin/Git Bash ship their own `whoami` ahead of Windows' and a bare call silently degrades the check to "cannot tell" — a permanent boot warning plus a hardening call that quietly does nothing. I hit that while testing, not by reading.

hristo2612:
Re-verified at head `40118b50` against current `main`. **Both blockers still stand**, and there's now a third item created by the merges below. Everything else in this PR is done well — see the credit section.

## Merged around you

`#97`, `#96`, `#100` and `#95` are now on `main`. **Please rebase.** This PR is not stacked on #97's tip — it carries #97's first commit but independently duplicates 10+ of its files (`build.mjs`, `posix-mode.ts`, `.gitattributes`, `tsconfig.build.json`, `vitest.config.ts`, `grok-interactive.ts`, and the same test call sites including `snapshot.test.ts`). After the rebase the diff should shrink to just `owner-only.ts`, `owner-only.test.ts`, `server.ts`, `pairing-challenge.ts` + its test, `auth-ux-api.test.ts`, `setup.ts` and `CHANGELOG.md`.

Note `snapshot.test.ts` changed on main in a way that matters to you: #95 removed the symlink materialization, so the assertion is now `expect(fs.existsSync(...AGENTS.md)).toBe(false)` alongside `expectPosixMode(...CLAUDE.md, 0o640)`. Don't reintroduce the `isSymbolicLink()` line.

## B1 — `enforceOwnerOnlyDirectory` reports success without verifying it (still open)

`owner-only.ts:161-169` runs one `icacls` invocation and unconditionally returns `true`. Per Microsoft's semantics, `/inheritance:r` removes **inherited** ACEs only, and `/grant:r <Sid>:<perm>` replaces previously granted explicit permissions **for the named trustee only**. A pre-existing **explicit** (non-inherited) ACE for an unrelated group is therefore neither dropped nor replaced — it survives, the function returns `true`, and `setup.ts:503` prints `✔ Restricted ~/.jinn to your account`.

That is a false assurance on the security-critical path, and it's the same "reports success, silently does nothing" class this PR exists to eliminate. Minimal fix, fails closed:

```ts
execFileSync(system32("icacls.exe"), [ /* … as today … */ ]);
// Verify rather than assume: /inheritance:r drops only INHERITED ACEs and
// /grant:r replaces only the trustees it names, so a pre-existing explicit ACE
// for an unrelated group survives. `undefined` (cannot tell) is a failure here.
return windowsPathIsOwnerOnly(directory) === true;
```

If you also want the call to *achieve* owner-only rather than just report it honestly, precede the grant with `icacls <dir> /reset` (drops explicit ACEs, leaving inherited ones for `/inheritance:r`). Either way keep the post-verify as the return value. Two small things in the same function: the enforce call sets no `stdio`, so `icacls` stderr prints raw over `jinn setup`'s formatted output (the read path already handles this at `:107`); and the POSIX branch could `return pathIsOwnerOnly(directory)` after the chmod for symmetry.

## B2 — `## [Unreleased]` and the release skill (still open, and I owe you a correction)

**My earlier framing was wrong**: I said this repo has "never used" `[Unreleased]`. It has — `5397119a` added one and the v0.18.0 release commit `e6cce7cf` correctly renamed it to `## [0.18.0] - 2026-06-04`. Apologies; there is precedent for handling it properly.

The hazard is still real. The release skill says to add the new entry *at the top* of `CHANGELOG.md` (which would orphan an `[Unreleased]` block below it), and extracts release notes with `awk '/^## \[0.X.Y\]/{flag=1; next} /^## \[/{flag=0} flag'`, which stops at the next `## [` and can never include `[Unreleased]`. So whether your 0755→0700 disclosure reaches the published notes depends on which of two undocumented behaviours the releaser picks. For a change that silently alters permissions on **every existing macOS and Linux instance**, that shouldn't be a coin flip. Simplest fix: name the concrete next version in the header.

## Not blocking, but worth doing while you're in here

- `server.ts:46-48` — `OWNER_ONLY_HEAL_MARKER` is declared *between* two import statements. Legal (imports hoist) but reads as an accident.
- Boot ordering: `ensureGatewayAuthToken(JINN_HOME)` runs at `server.ts:367`, the owner-only heal at `:398`. On a pre-existing `0755` home the token is materialized while the directory is still group-readable, then tightened. Tiny window, but moving the heal above `:367` costs nothing.
- `pairing-challenge.test.ts:84` calls bare `execFileSync("icacls", …)` — the exact PATH-shadowing hazard your own `system32()` helper exists to prevent, and documents at `owner-only.ts:35-40`.
- Same file, `:93-114`: the new `if (process.platform !== "win32") {` guard wraps ~20 lines that weren't re-indented.
- On Windows the proof check verifies DACLs but has no owner check corresponding to the POSIX `uid` triple, because `icacls /save` emits the DACL only. I don't think it's exploitable (planting the file needs write access to a home that is itself asserted owner-only), but the asymmetry deserves a sentence in the comment at `pairing-challenge.ts:46-52`.

## Credit

All four previously-required changes are properly done, and several are better than what I asked for: parsing SDDL trustees instead of localized `icacls` text is more robust than the SID-matching I suggested; resolving `whoami.exe`/`icacls.exe` through `system32()` closes a PATH-shadowing hole I hadn't raised; and the deliberate POSIX-heals / Windows-warns asymmetry with a marker file is a better answer to the "warning forever" objection than I expected. The #99 subsumption is correct — all three POSIX conditions are preserved byte-for-byte and Windows only *adds* a branch. Verified green: jinn-cli 284 files / 3495 passed, web 107 / 1133, typecheck clean.

## One request before merge

The `windows-ci` label is on this PR. Now that #97 has landed, the Windows leg can actually build — but note the root `postinstall` (`chmod +x … 2>/dev/null || true`) is still cmd.exe-incompatible and neither #97 nor this PR fixes it, so `pnpm install` will still fail there. `packages/jinn/scripts/fix-node-pty-permissions.mjs` (from #82) already does that job cross-platform and no-ops off Darwin — swapping the root `postinstall` to call it would be a one-line change and is the last thing standing between this repo and a green Windows run.

**None of this PR's Windows-specific logic — `icacls`, `whoami`, SDDL parsing, the ACL branch in the pairing test — has ever been executed**, by CI or by review. It's all POSIX-skipped or exercised through hand-written fixtures. Given this is the security-critical path, I'd like a green Windows run before merging.

e1010101:
Rebased and both blockers fixed. The Windows-run request is the interesting one — I measured it, and it costs more than one line.

## Rebase

Rebased onto current `main` with #97's commit dropped (`--onto origin/main c005b9fd`). The diff is now exactly the 8 files you predicted: `owner-only.ts` + its test, `server.ts`, `pairing-challenge.ts` + its test, `auth-ux-api.test.ts`, `setup.ts`, `CHANGELOG.md`. `snapshot.test.ts` is no longer in the diff at all, so #95's `expect(fs.existsSync(...AGENTS.md)).toBe(false)` stands untouched.

## B1 — verified, and it needed the `/reset` too

You were right on the mechanism, and taking `/inheritance:r` + `/grant:r` semantics seriously means the honest return value alone is not enough — without `/reset` the function would now correctly report failure where it previously reported false success, but still not achieve the thing. So both:

```ts
execFileSync(system32("icacls.exe"), [directory, "/reset", "/Q"], { ... });   // drops EXPLICIT ACEs
execFileSync(system32("icacls.exe"), [directory, "/inheritance:r", ...grants]); // drops INHERITED ones
return windowsPathIsOwnerOnly(directory) === true;                            // verify, never assume
```

`undefined` (ACL unreadable) counts as failure. POSIX returns `pathIsOwnerOnly(directory)` after the chmod for the same reason. Both enforce calls now pass `stdio` so `icacls` stops printing over `jinn setup`'s formatted output.

Tests: the return value must equal `pathIsOwnerOnly` afterwards, a missing path fails closed without throwing, and on POSIX a `0755` directory is actually tightened.

## B2 — `## [0.28.7]`

Named concretely rather than `[Unreleased]`, so it cannot depend on which of the two release behaviours the releaser hits.

## Boot ordering — this one was worse than "tiny window"

Fixed, and worth being precise about why it mattered: `ensureGatewayAuthToken` **creates** the token on first boot. On a pre-existing `0755` home the old order minted a fresh secret into a group-readable directory and only then tightened it, so the exposure covered the token's creation rather than merely its presence. The heal now runs first.

Also done: `OWNER_ONLY_HEAL_MARKER` moved out from between the imports; the pairing fixture resolves `icacls` from System32 via a local helper instead of bare `execFileSync("icacls", …)`; the POSIX-only guard body re-indented; and the DACL-vs-owner asymmetry stated at `proofOwnershipIsExclusive` — `icacls /save` emits the DACL only, so there is no Windows counterpart to the uid triple.

## The Windows logic has now been executed

You're right that none of it had been. It has now — I ran it against real ACLs on Windows 11, on a fresh `%TEMP%` subdirectory:

```
SID: S-1-5-21-…-1001
BEFORE ownerOnly: false
BEFORE trustees: …-1005 | …-2718948994 | …-1225693318 | …-2697887203 | …-1093152142
                 | …-3510416329 | S-1-15-3-…(AppContainer) | S-1-15-3-…(AppContainer)
                 | …-1001 | SY | BA          ← 11 trustees
enforce returned: true
AFTER  ownerOnly: true
AFTER  trustees:  BA | SY | …-1001            ← 3
CHILD  ownerOnly: true                        ← inheritance works
CHILD  trustees:  BA | SY | …-1001
```

Eleven trustees on a directory `mkdtemp` had just created — six unrelated account SIDs and two AppContainer SIDs. That is the case the pairing fixture's `enforceOwnerOnlyDirectory(home)` exists for, and it is a sharper illustration of the original finding than the PR body gives. `whoami.exe /user` parsing, SDDL trustee extraction, `/reset` + `/inheritance:r` + `/grant:r`, the post-verify, and child inheritance all executed on this path. `owner-only.test.ts`, `pairing-challenge.test.ts`, `auth-ux-api.test.ts` and `pair.test.ts`: 34 passed / 2 POSIX-skipped on Windows.

## A green Windows CI run is a bigger job than the postinstall

I measured the full suite on Windows 11 / Node 24 at current `main` (`24669ad0`): **17 failures across 11 files, 3485 passing.** The `postinstall` is real and your diagnosis is right, but it is not the last thing standing. What remains, all pre-existing on `main`:

| Failures | File(s) | Cause |
|---|---|---|
| 3 | `engine-limits.test.ts` | `writeCodexAppServerStub` writes a `#!/usr/bin/env node` file and chmods it; Windows has no shebang, so `spawn(bin, …)` never starts it |
| 4 | `mcp/__tests__/env-scrub.test.ts` | same class — "real spawn" of `scrub-entry.js` |
| 2 | `work-items/__tests__/migrate.test.ts` | multi-process SQLite contention (16/32 concurrent processes) |
| 2 | `sessions/__tests__/callback-concurrent-init.test.ts` | same |
| 2 | `work-items/__tests__/optimistic-concurrency.test.ts` | same, 30s timeouts |
| 1 | `pty-stream.test.ts` | 30s timeout pushing ~2,100 control sequences + 512KB through xterm-headless |
| **3** | `auth-ux-api.test.ts`, `pairing-challenge.test.ts` | **this PR fixes these** |

So this PR takes Windows from 17 to 14, and the remaining 14 are three distinct projects: shebang-executable test stubs (7), multi-process SQLite on Windows (6), and one slow xterm test. Adding a red Windows leg would also mean every subsequent PR shows a failing check.

The same `spawn` gap has a production edge worth separating out: `engine-limits.ts:355` uses bare `spawn(bin, ["app-server", "--stdio"])` and `:199` bare `execFile(bin, …)`. An npm-installed `codex` on Windows is `codex.cmd`, which Node refuses without a shell — the constraint `cli/skills.ts:145` and `setup.ts:673` already work around. Codex usage limits are therefore probably unavailable on Windows regardless of the tests.

My suggestion, and I'll take whichever you prefer: land this on the local-execution evidence above, and let me open the Windows CI work as its own PR — `postinstall` → `fix-node-pty-permissions.mjs`, the workflow leg, and the stub/timeout fixes — so the leg goes green the day it appears rather than arriving red. If you'd rather hold this PR until that lands, say so and I'll start on it now.

e1010101:
The Windows leg found a real bug in this PR, and it's mine. Treating it as blocking.

`pairing-challenge` ("accepts one exact owner-only proof file") and `auth-ux-api` ("mints a pairing code after one valid loopback filesystem challenge", 403 instead of 200) fail on the runner. Those are the three tests **this PR exists to fix**, so the owner-only path does not work on `windows-latest` even though it does on a local Windows 11 box. Note the negative cases all pass — `rejects expired and mismatched challenge proofs`, `rejects bearer-only pairing-code creation` — which is the signature of the check returning "not owner-only" when it should return the opposite.

My local evidence was real but insufficient: one machine is not the population, and you were right that the logic had never been executed anywhere else.

**I can't reproduce it locally**, so I've made the failure diagnose itself instead of guessing at a fix in security code. Pushed at `70ae3681`:

- `tempHome()` now *asserts* `enforceOwnerOnlyDirectory` succeeded rather than assuming it, and
- the acceptance case checks the proof file's own owner-only state separately from its POSIX mode.

Both messages carry the resolved SID and the actual trustee list, read on the failure path only. That splits the three candidate causes apart — unresolved identity (`sid=UNRESOLVED`), an ACL edit that didn't take (home not owner-only), or inheritance not reaching the proof file (home fine, file not) — and it's a permanent improvement rather than a debug commit, since a bare `expected false to be true` is close to undiagnosable on a machine nobody can log into.

**One ask:** fork PR runs sit at `action_required` until you approve them, so I can't trigger the Windows leg myself. If you approve a run on `70ae3681` I'll have the cause and a fix quickly. If you'd rather not spend approvals on it, tell me what the runner's `whoami /user` and a `icacls %TEMP%` look like and I may be able to work it out from that.

**Correction to my previous comment.** I said your Windows leg would face 14 pre-existing failures beyond this PR's. The real run says 6 failed files, and the multi-process SQLite tests that fail on my box pass on the runner — so my inventory was pessimistic and measured on the wrong machine. What the runner actually shows, besides mine: `engine-limits` (3, the shebang stubs), `performance.test.ts` (60s hook timeout), and two EPERM temp-cleanup failures in `browser-operator-authorization` and `workflow-provenance`. That's a materially smaller job than I described, and it makes the Windows CI follow-up more worth doing, not less.

Also worth flagging: the earlier run on `40118b50` that passed had no Windows leg at all — just `typecheck`, `unit-tests`, `build`. So nothing regressed between those two runs; the matrix is simply new, and this is the first time any of this code has run on Windows CI.

e1010101:
Rebased onto `main` at `ba4cd3df` (v0.28.7). One conflict, in `CHANGELOG.md`, and it needed a decision rather than a merge.

0.28.7 shipped while this was open, and its notes already cover #95/#96/#97/#100 — including "No-op migrations work on filesystems without symlink support" and "Windows source builds and port checks are reliable". This PR is not in that release, so its `## [0.28.7]` header (which I had just added at your request, to replace `[Unreleased]`) now collided with a released section describing different changes.

I moved the entry to a new **`## [0.28.8]`** above the released 0.28.7 rather than folding it into it. Naming a released version for an unshipped change would misattribute it, and the permission change affecting every existing POSIX instance is precisely the one that should not be filed under a version users have already installed. Rename it if 0.29.0 or another number is the real target — the point of naming a concrete version was to keep it out of the awk range's blind spot, and any concrete number does that.

Nothing else conflicted; the code commits replayed cleanly. Typecheck clean, and `pairing-challenge` + `auth-ux-api` + `owner-only` + `pair` + `migrations` are 53 passed / 2 POSIX-skipped locally.

The Windows question from my previous comment still stands: the leg fails on the runner for this PR's own tests, `70ae3681`'s self-diagnosis is now `39bd3f17`, and fork runs need your approval before it can tell us why.

e1010101:
Thanks for approving the run — it named the cause immediately, and it is a genuine bug in this PR that is **not confined to CI**.

## What the runner reported

```
could not establish an owner-only home at C:\Users\RUNNER~1\...\jinn-pair-challenge-bLUZDg
  — sid=S-1-5-21-1742564184-1656218818-310408600-500 trustees=[BA, SY, LA]
```

The ACL was **exactly right** — Administrators, SYSTEM, and the user, nothing else. The verification could not recognise its own grant.

`windows-latest` runs as the built-in Administrator (RID **500**), and SDDL serializes some well-known SIDs as two-letter aliases. So `icacls /grant:r *S-1-5-21-…-500` is written back by `icacls /save` as `LA`, not as the raw SID it was given. `windowsPathIsOwnerOnly` compared trustees against the raw SID only, so it read the user's own ACE as a stranger's and answered "shared".

## Why this matters beyond the test suite

On any machine whose operator is the built-in Administrator — not rare on Windows — `pathIsOwnerOnly(JINN_HOME)` answers false for a directory that *is* owner-only. That means `server.ts` warns "readable beyond your account, the gateway token and secrets are exposed" on **every boot**, about an exposure that does not exist, and `jinn setup`'s hardening reports failure immediately after succeeding. A security check that cries wolf on a correctly configured machine is worse than one that is merely quiet.

It also means the bug predates this round: the `t === mySid` comparison has been in the PR from the start. It only became *detectable* when the previous commit made `enforceOwnerOnlyDirectory` verify its work — while it returned `true` unconditionally there was nothing to disagree with it. Your B1 found a second bug by fixing the first.

## Fix (`918bcae9`)

The current user's SID is now matched against every SDDL spelling of that exact account, derived from the RID rather than accepted blindly:

- RID 500 → also `LA`; RID 501 → also `LG`; domain accounts (RID ≥ 1000) have no alias.
- `LA` denotes **this machine's** Administrator, so it counts as the current user only when the current user is that account. For anybody else it is a different principal and still reads as shared — pinned by a test both ways.

The comparison moved out of the icacls path into an exported pure predicate, `trusteesAreOwnerOnly(trustees, mySid)`, so the verdict is checkable without a Windows box to read an ACL from — same reasoning as `parseSddlTrustees`. The runner's exact ACL (`["BA","SY","LA"]` against the RID-500 SID) is a fixture, alongside rejection of `WD`/`AU`/`BU` and of another account's alias.

Local: 39 passed / 2 POSIX-skipped across `owner-only`, `pairing-challenge`, `auth-ux-api`, `pair`; typecheck clean.

**Could I trouble you for one more approval on `918bcae9`?** I expect the three failures in this PR's tests to clear. The other three files on that leg (`engine-limits` ×3, `performance.test.ts`, and the two EPERM temp-cleanup suites) are pre-existing on `main` and not this PR's.

The self-diagnosing assertions stay in. They cost one `icacls` call on the failure path and turned an undiagnosable "expected false to be true" into a one-line root cause on a machine I cannot log into.

e1010101:
Confirmed on the runner: **this PR's three failures are gone.** `pairing-challenge.test.ts` and `auth-ux-api.test.ts` both pass on `windows-latest` at `918bcae9`. The `LA` alias was the whole of it.

The leg went from 6 failed files to 4, and none of the remainder belongs to this PR:

| File | Cause |
|---|---|
| `shared/__tests__/engine-limits.test.ts` (3) | `writeCodexAppServerStub` writes a `#!/usr/bin/env node` file; Windows has no shebang, so `spawn(bin, …)` never starts it |
| `activity/__tests__/performance.test.ts` | `Hook timed out in 60000ms` — a benchmark fixture on a slower runner |
| `gateway/__tests__/browser-operator-authorization.test.ts` | `EPERM … syscall: 'rm'` removing the vitest temp root |
| `sessions/__tests__/workflow-provenance.test.ts` | same EPERM teardown |

All four reproduce on `main`, so the security change is now clean on both platforms — but the leg is still red, and you asked for green before merging.

I'd like to take those four as the separate Windows-CI PR I proposed: the `postinstall` swap to `fix-node-pty-permissions.mjs`, a teardown that tolerates Windows' delayed handle release instead of failing the suite, the benchmark's hook budget, and cross-platform stubs (or an explicit platform skip where a shebang genuinely cannot be expressed — the approach #97 already established). That keeps unrelated infrastructure out of a security change, and once it lands this PR rebases onto a green leg.

Two things I'd need from you for that: the `windows-ci` label on the new PR so its leg actually runs, and a run approval, since fork PRs sit at `action_required`.

Say the word and I'll open it. If you'd rather I fold the four into this PR to unblock it sooner, I can do that instead — just tell me which you prefer.

hristo2612:
Merged, along with #94. Genuinely excellent work across all six of these, and the RID-500 `LA` alias catch is the kind of finding that would have shipped as a permanent false security warning to a whole class of Windows users if you had stopped at "CI is red for unrelated reasons" — thank you.

hristo2612:
Yes to the Windows CI PR, please go ahead.

Scope as you proposed: the workflow leg from opt-in to always-on, plus the six remaining failures. One correction on the list, from the runner rather than from prediction: the root `postinstall` is **not** a blocker. On run 30317780975 the windows-latest job's `pnpm install --frozen-lockfile` and `pnpm build` steps both come back `success`, and only `pnpm test` fails. I predicted that postinstall would break install, twice, and the runner disagreed with me both times. So the job is smaller than either of us described: the three shebang stubs in `engine-limits.test.ts`, the two EPERM temp-cleanup suites, and the `performance.test.ts` timeout. Swapping the root `postinstall` to `fix-node-pty-permissions.mjs` is still worth doing as tidiness, just not as a prerequisite.

Order matters more than speed here: green first, then flip the leg on. An always-on leg that arrives red trains everyone to ignore it, which is worse than the opt-in we have now.

I've also filed the `.cmd` spawn gap you separated out, as #103. That one is yours if you want it, but no obligation, and it is genuinely independent of the CI work.

On the switch-flipping problem: you are right that you shouldn't need me for this, and the approval gate is already gone (nothing has sat at `action_required` since 26 Jul). What's left is that only a maintainer can apply the `windows-ci` label. Since your PR makes the leg always-on, that stops mattering, so I'd rather fix it by landing your PR than by working around it. Ping me on this thread if you need the leg on a branch before then and I'll label it same day.

PR #127, opened 2026-08-04
feat(docker): run the gateway in a container

yoobi:
## Why

The engine spawns `claude` with `--dangerously-skip-permissions` on every turn, so
the approval gate is off for everything that process can reach. On a host install
that is a whole home directory; in a container it is only the paths you chose to
mount. `docker-compose.yml` keeps that list in one place and labels it the blast
radius.

## What it adds

`Dockerfile`, `docker-compose.yml`, an entrypoint, a HEALTHCHECK, and
`docs/docker.md`. The build is multi-stage and runs the ROOT `pnpm build`, which
also syncs the dashboard the gateway serves — filtering to the jinn package yields
a gateway with no UI. `pnpm deploy` then ships a runtime with no build toolchain
in it: 215MB against the 650MB of workspace tree it replaces. `claude-code` is
pinned by build arg with its auto-updater off, because engine drift is a realised
failure mode.

The published port is loopback-only, and the gateway refuses to serve a network
bind without auth, so first-run ends in `jinn pair`.

## Three seams it exposed, fixed for every install

- `seedTrust` now records the Bypass Permissions consent that global onboarding
  implied up to claude-code 2.1.170 and no longer does. Without it every turn hangs
  on a dialog no PTY can answer.
- Transcript reads, the Claude auth guard and the engine child env follow
  `CLAUDE_CONFIG_DIR`; five call sites hardcoded `~/.claude`.
- Stale pids are keyed on hostname plus boot identity. A container restarts pids
  from 1, so the reaper was SIGTERMing unrelated live processes — possibly PID 1.

## The invariant worth reviewing

`JINN_HOST`/`JINN_PORT` describe this process's binding, and `config.yaml` sits on
a volume that outlives the container. `loadConfig()` resolves them into every
config it returns, so the hazard is a writer persisting them: a home later opened
on a workstation would silently bind every interface. `saveConfigAtomic` restores
the file's own values, so no writer can leak them — that is the single place to
check if you add another.

Related consolidation: "where is this gateway reachable" is one resolver instead of
five copies that had drifted apart, and the gateway records its own URL so the
shell healthcheck does not re-derive wildcard and IPv6 handling.

## Verification

The `docker` CI job builds the image, then loads better-sqlite3 and spawns a PTY —
neither of which `docker build` exercises. It boots the gateway on an empty volume
and asserts `/api/status`, the dashboard actually served at `/`, `HEALTHCHECK`
reaching healthy, and recovery from a stale `gateway.pid`.


yoobi:
@hristo2612 whenever you have the time. My proposition to support docker :)

PR #133, opened 2026-08-24
Add owner-only Telegram auth flows for Claude and Codex

ajdevy:
Implements the reviewed Telegram authentication flow for the owner.

- Adds /auth status, /auth claude, /auth codex, and bounded input/cancel handling.
- Prompts only in the owner private chat after provider authentication failures.
- Keeps auth links/codes out of normal sessions and logs.
- Adds the pinned Codex CLI image integration and tests.

Verification: focused Telegram auth tests 40/40 passed; full pnpm typecheck passed.

PR #134, opened 2026-08-24
Fix owner-only Telegram provider authentication

ajdevy:
## Summary
- add owner-only Telegram `/auth status`, `/auth claude`, and `/auth codex` flow
- show a login prompt after Claude/Codex authentication failures instead of returning only the raw error
- keep provider credentials and one-time codes out of ordinary Telegram replies
- install and verify pinned Claude/Codex CLIs in the Jinn image

## Verification
- `pnpm footguns`
- `pnpm ratchet --check`
- `pnpm --filter jinn-cli migration:check -- --base-ref v0.31.0 --version 0.31.0`
- focused Telegram tests: 32 passed
- `pnpm --dir packages/jinn exec tsc --noEmit --pretty false`

This supersedes the earlier stale-base PR #133; the branch is based on the current upstream `main`.


hristo2612:
Consolidating on #143, which is a rewrite of this branch. I've left the full review there. Could you close this one so we're reviewing a single branch?

Two things from here worth carrying **into** #143, because #143 regressed them:

1. **`codex login status`** is the right Codex check. #143 replaced it with `fs.existsSync(".codex/auth.json")`, which can't distinguish a valid session from an expired one.
2. **Gating on `config.telegramAuth?.enabled === true` in the constructor.** #143 always constructs its integration, which reads state from disk and can spin up a second `TelegramBot` on the same token even when the feature is off.

One thing that must **not** carry over, and it's the reason this branch couldn't merge as-is:

**The migration bundle edits are unsafe.** This adds an identical `docs/connectors.md` record to both `template/migrations/0.31.0/manifest.json` and `0.32.0/manifest.json`, with the same `baseSha256` (`b439377e…`). Both versions are already tagged and released.

A user upgrading 0.30.1 → 0.32.0 applies the record twice. The first pass moves the file to `7400864e…`; the second pass still expects base `b439377e…`, mismatches, and conflicts. Migration bundles for shipped versions are history and shouldn't be rewritten. New docs belong in one record against the next unreleased version.

Related: the root-level `docs/connectors.md` this adds isn't a path that exists in the repo, and its contents are a 40-line fragment rather than the real document, which lives at `packages/jinn/template/docs/connectors.md`. Looks like an artifact rather than something intended.

Also please leave `scripts/upgrade-lab/assertions.mjs` and the connector's send/typing/edit path alone in the next round. Both changes are behaviour-neutral and unrelated to Telegram auth, and they make the diff harder to review than the feature warrants.

ajdevy:
Superseded by PR #143, where the review feedback has been consolidated and implemented. Closing this duplicate PR so review and merge can proceed from the current head.

PR #143, opened 2026-08-25
fix(telegram): harden owner authentication flows

ajdevy:
Reviewed owner-only Telegram authentication flows for Claude/Codex.

Local verification on upstream main base:
- focused Vitest: 87 passed
- Jinn build: passed
- TypeScript typecheck: passed
- targeted ESLint: passed
- git diff --check: passed

The change keeps secret-bearing input deletion, fail-closed provider verification, timeout retry guidance, lifecycle-safe listener cleanup, and owner-only menu configuration.


hristo2612:
Thanks for this, and for the care that went into it. The security thinking is the strongest part: owner allowlist, private-chat-only, fail-closed on non-owners, scrubbing secret-bearing input (including from non-owners, which #134 missed), never echoing the code back. The tests are real integration tests with fake timers and filesystem state, not mock theatre. The CLI invocations are correct, I verified `claude auth login --claudeai`, `claude auth status --json`, and `codex login --device-auth` all behave as you assume.

I do want this feature. I don't use Telegram myself, but other Jinn users do, and provider re-auth from the phone is a real gap.

What I can't merge is the current size and shape. 3,158 added lines for "log in to Claude/Codex from Telegram" is roughly 5x what the problem needs, and a few of the decisions will actively break for anyone not running the Docker image. Below is what I need changed and the architecture I want to end up with.

---

## 1. It has to be optional by design

This is the hard requirement. Jinn ships Telegram to everyone; only a few will enable this.

With `telegramAuth` absent or `enabled: false`, the connector must behave **byte-identically to today**:

- nothing constructed, no `TelegramAuthIntegration`, no `TelegramAuthMenu`
- no filesystem read (right now `TelegramAuthMenuState`'s constructor reads state on every Telegram connector boot, enabled or not)
- no second `TelegramBot` instance on the same token
- no per-message work (`ensureMenu()` currently runs on every inbound message)
- `/auth ...` falls through to the agent exactly as it does now

Gate on `config.telegramAuth?.enabled === true` in the constructor and build nothing otherwise. #134 actually got this right.

## 2. It is currently Docker-only, and nothing says so

`/home/node` is hardcoded in four places across `auth-flow-manager.ts`, `telegram-auth-integration.ts`, and the Codex probe. There is no precedent for that anywhere in `packages/jinn/src`. The gateway's primary install path is npm/Homebrew running natively.

On my own machine:

```
$ which claude
/Users/…/.local/bin/claude

$ env -i PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin claude auth status --json
env: claude: No such file or directory
```

So on a native install: Claude status always reports "not authenticated", Codex status always reports "not authenticated" (`fs.existsSync("/home/node/.codex/auth.json")`), and the PTY spawn fails on a cwd that doesn't exist. Silently.

Inherit `process.env`. Don't override `HOME`. Pass `CLAUDE_CONFIG_DIR` / `CODEX_HOME` through only if they're already set, otherwise let the CLI pick its own defaults. Use `os.homedir()` or the gateway cwd. The Dockerfile already sets `CLAUDE_CONFIG_DIR`, and your `ENV CODEX_HOME` addition covers the container case correctly, so inheriting is strictly better than hardcoding.

## 3. Ask the CLI, don't guess

`fs.existsSync(".codex/auth.json")` can't tell a valid session from an expired one. #134 had this right with `codex login status`, exit code 0. Go back to that. `execFile` with a `timeout` option is enough for both providers; no manual SIGTERM/SIGKILL ladder needed.

## 4. Drop the menu persistence subsystem entirely

`telegram-auth-menu.ts` + `telegram-auth-menu-state.ts` + their tests are roughly 800 lines to make four commands appear in Telegram's `/` autocomplete. They contain SHA-256-derived state filenames, atomic tmp+rename writes at mode 0600, corrupt-file archival to `.corrupt-<timestamp>` copies that are never cleaned up, exponential backoff to a 15-minute ceiling, lifecycle and configuration generation counters, and stale-owner reconciliation driven by regex-matching Telegram's English error strings.

`setMyCommands` is idempotent and free. All of that machinery exists to avoid re-issuing a call that costs nothing.

If you want the `/` menu, and I think it's nice to have:

```ts
try {
  await bot.setMyCommands(commands, { scope: { type: "chat", chat_id: ownerId } })
} catch (err) {
  logger.warn(`[telegram] could not publish auth command menu: ${err}`)
}
```

Once, on start, per owner. No retries, no persistence, no new state files under `~/.jinn/`. If an owner is removed from config, their stale menu is a cosmetic wart, not something worth a reconciliation protocol.

Also please drop `setTelegramAuthMenuStateScope()`. A mutable module-level global that exists only so tests can redirect a path doesn't belong in production code.

## 5. Remove the dead code

`redactAuthOutput` is never called in production. Its only caller is its own test. Either wire it into the path that needs it or delete it. Same in #134, so it looks inherited.

## 6. Don't game the size ratchet

`index.ts` gains ~12 lines of real code and loses 11 blank lines so `size-baseline.json` can report 459 → 458. `shared/types.ts` gets the same treatment, and `telegram-config-types.ts` is a new 5-line file holding one interface so that `types.ts` doesn't grow.

The ratchet's message says "split the file by responsibility." Whitespace deletion satisfies the number and not the rule, and it adds diff noise in files the feature has no business touching. Please revert the blank-line churn. If the auth entry point genuinely doesn't fit in `index.ts`'s budget, one focused module is the right answer, but it should be an actual seam, not a container for a type alias.

## 7. Redundant work in the PTY hot path

`attach()` calls `discoveryMessageLines(flow)` on every chunk, which re-runs `extractDiscovery` over the full 64 KB output buffer, while `updateDiscovery` already extracted from a 4 KB tail on the same chunk. An interactive CLI with a spinner emits many chunks per second for up to ten minutes. Keep the tail-based extraction and drop the full-buffer pass.

Minor, same area: `appendBoundedText` slices a UTF-8 buffer at an arbitrary byte offset and can split a multi-byte character.

## 8. Docs are missing

This adds a user-facing config key and four bot commands with no documentation. Please add a section to `packages/jinn/template/docs/connectors.md` covering `telegramAuth` (`enabled`, `ownerUserIds`, `flowTtlSeconds`), the commands, the private-chat-only rule, and the fact that owner IDs must also be in `allowFrom`.

One migration record, for the next **unreleased** version only. Please don't touch `0.31.0` or `0.32.0`; both are tagged and shipped. (See my note on #134 for why that breaks upgrades.)

---

## The architecture I'm after

Roughly this, and closer to 250 lines of production code than 1,000:

```
connectors/telegram/
  auth.ts           # parse commands, own the flow lifecycle, the whole public surface
  auth-providers.ts # the only place that knows "claude" from "codex"
  __tests__/auth.test.ts
```

**`auth-providers.ts`** is a table, not a set of ternaries:

```ts
export const PROVIDERS = {
  claude: {
    label: "Claude",
    login: ["claude", ["auth", "login", "--claudeai"]],
    status: async () => JSON.parse(await run("claude", ["auth", "status", "--json"])).loggedIn === true,
  },
  codex: {
    label: "Codex",
    login: ["codex", ["login", "--device-auth"]],
    status: async () => (await runExit("codex", ["login", "status"])) === 0,
  },
} as const
```

Right now `provider === "claude" ? … : …` appears in six places across four files. Adding a third provider should mean one entry here and nothing else.

**`auth.ts`** owns: command parsing, the owner set, the active-flow map, PTY spawn/attach/teardown, and status. One class. The `AuthFlowManager` → `AuthFlowProcessManager` → `ProviderStatusVerifier` chain (and #134's `AuthFlowRuntime`, a class that only forwards) is more indirection than a four-command feature can pay for. `ProviderStatusVerifier`'s generation counters and 5-second result cache are solving a contention problem that two commands and one owner do not have.

**Keep exactly as-is**, this is the part I want preserved verbatim:

- owner IDs intersected with `allowFrom`, with the warning when they diverge
- private chat only
- delete secret-bearing input, including from non-owners, and warn when deletion fails
- codes matched against a strict short-code pattern; tokens never accepted
- fail closed and stay silent toward non-owners
- the underscore command forms (`/auth_claude`) so Telegram autocomplete works, though a single generic `^/auth[_a-z-]*` prefix guard would replace the enumerated token synonyms

**Don't touch** anything the feature doesn't require. No refactor of the send/typing/edit path, no unrelated churn in `scripts/`.

---

## Acceptance criteria

- [ ] With `telegramAuth` absent, `git diff` of connector behaviour is nil: no state files, no extra bot instance, no per-message work, `/auth` reaches the agent
- [ ] No absolute container paths in `packages/jinn/src`; verified working on a native (non-Docker) install
- [ ] Codex status comes from `codex login status`
- [ ] Menu publication is one call per owner on start, no persistence, no retries
- [ ] No dead exports
- [ ] No blank-line-only edits, `size-baseline.json` reflects real structure
- [ ] `telegramAuth` documented; one migration record against the next unreleased version
- [ ] The security tests survive the shrink

Happy to look at an outline before you rewrite if that saves you a round. And please close #134 in favour of this one so we're reviewing a single branch.

ajdevy:
Conflicts with `main` are resolved in `9a67555c9`. Coverage is serialized to prevent the shared SQLite race, and Windows shell-contract paths are normalized. VPS verification passed (587 files, 5985 tests, 7 skipped; typecheck, lint, ratchet, migration, Docker checks), and GitHub CI `33222913082` plus migration `33222913086` are green. Ready to merge.

hristo2612:
This is a big improvement. +3,158 → +1,222, 29 → 21 files, and I verified all eight criteria against the tree at `03ad3704` rather than the checklist:

- No `/home/node` anywhere under `connectors/telegram/`; `runCommand` inherits `process.env` and `process.cwd()` ✅
- `codex login status` by exit code ✅
- `redactAuthOutput` and the menu state/manager modules are gone ✅
- `0.31.0` / `0.32.0` untouched; only a `0.33.0` bundle ✅
- `size-baseline.json` now reports 459 → **480** and types.ts 833 → **838**. Raising the budget honestly is exactly right, and the blank-line churn is gone (11 → 1) ✅
- Menu is one owner-scoped `setMyCommands`, fire-and-forget, warn on failure ✅

`auth-providers.ts` is the shape I was after. `acceptedInputSources` / `invalidInputMessage` on the provider record is a better idea than what I sketched, since it keeps the Claude callback quirk with the provider that has it. The `jinn-codex` volume in `docker-compose.yml` is a real catch I didn't ask for.

One blocker before this can merge.

---

## The auth handler now runs before the `allowFrom` gate, and bare tokens are treated as auth input

Two changes combine into a behaviour bug.

**1.** `index.ts` moved the auth call ahead of the allowlist check:

```ts
const userId = telegramMsg.from?.id;
if (this.auth && await this.auth.handleIncoming(...)) return;   // <-- now first

if (this.allowedUsers) {
  if (userId === undefined || !this.allowedUsers.has(userId)) return;   // <-- was first
}
```

In the previous revision `allowFrom` was the outer perimeter and auth ran inside it.

**2.** `parseAuthInput` accepts a **bare** token, with `CODE_PATTERN = /^[A-Z0-9](?:[A-Z0-9-]{2,30}[A-Z0-9])$/`. That is any uppercase alphanumeric/hyphen string of 4 to 32 characters. `sensitive` is then true for it, and `scrub()` deletes unconditionally, with no owner check:

```ts
private async scrub(message, sensitive, owner) {
  if (!sensitive || message.messageId === undefined) return "";
  try { await this.deleteMessage(message.chatId, message.messageId); return "" }
  ...
}
```

So with `telegramAuth.enabled: true`:

- **A user explicitly excluded by `allowFrom`** who sends `TESTING`, `README`, `ACME-CORP`, `BTC-USD` has that message **deleted by the bot** and swallowed. The bot now acts on behalf of people the allowlist exists to exclude.
- **An owner** who types `HELLO-WORLD` in ordinary conversation, with no auth flow running, gets the message deleted and a reply of `"No authentication flow is active."` (`auth.ts:213`). It never reaches the agent. In a group where the bot is admin, that applies to everyone.

`resolveOwnerIds` does intersect owners with `allowFrom`, so a stranger can never become an owner. That part is sound. But `scrub()` runs before ownership matters, so the deletion happens anyway.

**Fix, two lines of intent:**

1. Move `handleIncoming` back below the `allowFrom` check. Nothing in the auth path needs to see a non-allowlisted user.
2. Only treat a bare token as auth input when that owner has an active flow. Something like an early `if (this.active.size === 0 && !AUTH_PAYLOAD_PATTERN.test(raw)) return false` at the top of `handle()`, so ordinary conversation is never captured.

Keep the explicit `/auth … <payload>` scrub unconditional. That is the deliberate hardening from the last round and it should stay. It's only the **bare** token path that needs the active-flow gate.

I like the bare-code convenience, it matches what your `instructions` strings promise. It just has to be scoped to a live flow.

---

## Two small things, not blockers

- **Drop the version bump.** `packages/jinn/package.json` 0.32.0 → 0.33.0 is a maintainer action tied to the release process, and it will conflict. Please revert it and leave the `0.33.0` migration bundle in place; I'll bump when I cut the release.
- **`format.ts` is out of scope**, though I understand why: `stripTelegramMarkdown` was eating `_` and `*` in the auth URL. The `\x00URL<n>\x00` placeholder is correct as written (the module-level `/g` regex is safe with `replace`, which resets `lastIndex`). Worth a one-line comment saying it exists so a login link survives the plain-text fallback, otherwise it reads as unrelated.

A regression test for the two cases above would be good to have alongside `telegram-auth-optional.test.ts`: a non-allowlisted sender and an owner with no active flow, both sending a bare uppercase token, asserting the message is neither deleted nor consumed.

CI is green across all ten checks. Once the ordering is fixed I'm happy to merge this.

hristo2612:
I put this head through four independent review passes (security, lifecycle/concurrency, integration/regression, parsing) with instructions to prove every claim by running it rather than reading it. Below is what survived filtering. I dropped a dozen findings that were style, unreachable, or theoretical.

The structure of the PR is good and I'm not asking you to change it. Every issue here is a bug inside the shape we already agreed on.

---

## 1. BLOCKER — `extractDiscovery` treats its first partial match as final, so the code and URL you send are usually wrong

This is one root cause with five symptoms, and it is the reason I think the happy path does not currently work.

`extractDiscovery` is a best-effort scraper run against an **incomplete** buffer, but `auth.ts:168-169` latches its first non-empty result permanently. Nothing distinguishes "matched at end-of-buffer, may be truncated" from "matched with a terminator after it". `DISCOVERY_CODE_PATTERN`'s `(?![A-Z0-9-])` lookahead is satisfied by end-of-buffer, and the URL pattern `[^\s"'<>\`]+` likewise just runs to the end of what has arrived.

**a) Truncated URL.** Real 101-char Claude OAuth URL, cut at a chunk boundary:

```
chunk1: "Browser did not open. Visit:\nhttps://claude.ai/oauth/authorize?client_id=9d1c250a-e6"
chunk2: "1b&response_type=code&state=hK2mQ9xLpR4vN8wT\n"

sent to user: "Continue authentication:\nhttps://claude.ai/oauth/authorize?client_id=9d1c250a-e6"
complete URL ever sent: false
```

**90 of 99** chunk-boundary positions in that line latch a truncated URL. Only a boundary landing after the terminating newline is safe. Real OAuth URLs are ~200-250 chars, node-pty reads are chunked, and the TUI wraps at your `cols: 120`. This is the common case, not the edge.

**b) Truncated device code.** Real code `WDJB-MJHT`:

```
chunks: "Enter code: WDJB" + "-MJHT\n"
sent  : "Continue authentication:\nDevice code: WDJB"
split@18 -> "WDJB-M"   split@19 -> "WDJB-MJ"   split@20 -> "WDJB-MJH"
```

**c) Any uppercase banner line becomes the code.** `auth-support.ts:124`'s fallback is `split("\n").find(isAuthCode)` with no keyword context at all:

```
chunk1: "\nWARNING\nContacting authorization server...\n"
chunk2: "\nYour device code is: WDJB-MJHT\n..."

sent: "Device code: WARNING"
real code WDJB-MJHT ever sent: false
```

`ERROR`, `STARTING`, `WAITING`, `CONNECTING` all do this.

**d) `code=` in the auth URL's own query string is harvested as the device code.** `DISCOVERY_CODE_PATTERN`'s bare `code` alternative has no word boundary and no URL-context exclusion:

```
"Open https://claude.ai/oauth/authorize?code=A1B2C3D4E5&state=xyz"
-> {url: "...", code: "A1B2C3D4E5"}
```

The user is told to enter the OAuth response parameter as their device code.

**e) First `https://` in the buffer wins**, so a banner link beats the auth link: `"Claude Code v2.1.0 - docs: https://docs.claude.com/cli\nSign in at https://claude.ai/oauth/..."` yields the docs URL.

**Fix for all five:** require a real terminator (whitespace or newline, not end-of-buffer) before accepting a URL or a code, and let a later longer match supersede an earlier one until the process exits. Drop the context-free bare-line fallback in (c), and anchor the code match to an actual label so (d) stops firing inside query strings.

**This needs a chunk-boundary test.** Feed the same output split at every byte position and assert the delivered value is always the complete one. CI is green today because nothing exercises a split.

---

## 2. BLOCKER — `CODE_PATTERN` matches any bare uppercase token, and one consequence is keystroke injection into the live login PTY

I flagged the ordering half of this last round. It is worse than I described.

`CODE_PATTERN = /^[A-Z0-9](?:[A-Z0-9-]{2,30}[A-Z0-9])$/` matches any 4-32 char uppercase/digit/hyphen token. Confirmed matching: `DONE`, `TODO`, `ASAP`, `BLOCKED`, `LGTM`, `SHIP-IT`, `README`, `PLA-269`, `ICI-548`, `JIN-42`, `2026-08-27`. Those last four are exactly the traffic this project runs through chat.

Three distinct outcomes:

**a) Owner, flow active — the message is typed into the provider's TTY.** `auth.ts:217` runs `flows[0].pty.write("LGTM\r")`. Verified: `bytes written to provider PTY: ["LGTM\r"]`. An owner who starts `/auth_claude` and then says anything in caps has it injected as keystrokes into the interactive login prompt.

**b) Owner, no flow — the message is destroyed.** `scrub()` deletes it, `handleInput` returns `true`, `index.ts:105` returns, and the agent never sees it. The owner gets the misleading `"No authentication flow is active."` Verified: `{"case":"owner-private-DONE","swallowed":true,"deleted":[[10,5]]}`.

**c) Non-allowlisted user, any chat — deleted silently, no reply.** `handleIncoming` still runs before the `allowFrom` gate, and `scrub()` has no owner check, so the delete happens before the `chatType !== "private"` guard too. Verified in a group: `{"case":"nonowner-group-PLA-269","swallowed":true,"sent":[],"deleted":[[-100,6]]}`.

**Fix:** move `handleIncoming` below the `allowFrom` check, and only evaluate a bare token as auth input when that specific owner has an active flow. Keep the explicit `/auth … <payload>` scrub unconditional. The bare-code convenience is good and matches your `instructions` strings; it just has to be scoped to a live flow.

---

## 3. BLOCKER — the `format.ts` URL guard breaks markdown stripping for every Telegram user, feature enabled or not

`URL_PATTERN = /https?:\/\/[^\s"'<>]+/gi` does not exclude `*`, `_` or `~`. When emphasis **ends** with a URL, the closing delimiter is absorbed into the placeholder and the opener is orphaned in the plain-text fallback:

```
in : "**Release notes: https://github.com/jinn/jinn/releases/tag/v0.33.0**"
old: "Release notes: https://github.com/jinn/jinn/releases/tag/v0.33.0"
new: "*Release notes: https://github.com/jinn/jinn/releases/tag/v0.33.0*"
```

With two such spans the surviving openers pair across each other and the closers get glued onto the URLs, corrupting the links:

```
in : "*x https://a.com* and *y https://b.com*"
old: "x https://a.com and y https://b.com"
new: "x https://a.com* and y https://b.com*"
```

This is on the shared send path (`index.ts:372`) and has nothing to do with `telegramAuth`, so it hits every Telegram user on upgrade. The underlying fix you were going for is real and does work for `https://example.com/path_with_underscores`; only the delimiter-adjacent case regresses. Exclude trailing `*_~` from the URL match, or strip them off the captured URL and re-emit them.

While you are on that line: `stripped.replace(/\x00URL(\d+)\x00/g, (_m, i) => urls[Number(i)])` has no bounds check, so a literal `\x00URL0\x00` in message text becomes the string `"undefined"`. Fall back to the matched text.

---

## 4. HIGH — the callback-URL path fails open, so an unparsed OAuth code is forwarded to the model instead of being scrubbed

`auth-providers.ts:61` tells the owner "If Claude redirects to a localhost /callback URL, send the full URL here." If `parseClaudeCallbackCode` does not recognise what they paste, `parseAuthInput` returns `null`, a bare URL does not match `AUTH_PAYLOAD_PATTERN`, so `sensitive` is `false`, `handle()` returns `false`, and the message falls through `index.ts:105` into the normal handler.

The live authorization code and state are then persisted in the session store, sent to the model provider as ordinary conversation text, and left in Telegram history undeleted.

The specific gap that makes this reachable: `CLAUDE_CALLBACK_CODE_PATTERN = /^[A-Za-z0-9]{40,128}$/` rejects `-` and `_`, while the adjacent `CALLBACK_STATE_PATTERN` explicitly allows both. The secret half has the stricter charset than the non-secret half, which is backwards, and base64url codes contain exactly those characters.

I want to be precise about what is proven: the fail-open **shape** is confirmed by reading the control flow, and a fourth reviewer separately confirmed the URL host/scheme/path accept set is tight and correct. What I have not empirically confirmed is the exact charset Anthropic issues today. That does not really matter for the fix. Any input the owner was told to paste, which then fails to parse, should be treated as sensitive and scrubbed rather than forwarded. Fail closed on anything reaching a `/callback` shape.

---

## 5. MEDIUM — `stop()` does not latch, so a flow started across the `scrub` await outlives connector shutdown

`AUTH_PAYLOAD_PATTERN` matches the legacy `/auth claude` form, so that command is `sensitive` and `handle()` awaits a real Telegram `deleteMessage` round-trip (~100-500 ms) before dispatching.

1. Owner sends `/auth claude`; `handle()` is parked on `await this.scrub(...)`.
2. Config hot-reload fires. `reloadConnectorRegistry` (`gateway/server.ts:345-355`) calls `TelegramConnector.stop()` → `auth.stop()`. `active` and `pending` are both empty, so it is a no-op.
3. `deleteMessage` resolves, `startFlow` spawns a `claude auth login` PTY, inserts into `active`, arms the 600 s timer.
4. That instance is already out of `connectorMap` and `stop()` will never be called on it again. The child process is never killed, the timer holds the event loop, and its `onData` keeps calling `sendMessage` on a bot that has stopped polling. A subsequent `/auth_claude` on the new instance spawns a competing login.

Verified: `spawn calls: 1  kill calls: 0  timers: 1` after `stop()`.

**Fix:** set a `stopped` flag in `stop()` and check it in `startFlow` after every await.

---

## Housekeeping

- **Revert the `0.33.0` version bump.** Beyond it being mine to make, it hard-fails the release: `.github/workflows/publish-npm.yml:74-77` refuses to publish when `CHANGELOG.md` has no `## [0.33.0]` section, and the PR adds none. Keep the 0.33.0 migration bundle, drop the `package.json` change.
- **`enabled: true` with no resolvable owners is purely destructive.** `ownerUserIds` absent, or listing an id not in `allowFrom` (silently dropped by `resolveOwnerIds` with only a log warning), leaves `this.owners` empty. No command can ever succeed, but the scrub and swallow paths in findings 1 and 2 stay fully live. Refuse to construct, or log an error, when `enabled` is true and no owner resolves.
- **The new Dockerfile assertion is tautological.** `docker-entrypoint.test.ts:151-158` asserts `toContain("/home/node/.codex")`, which the `ENV CODEX_HOME=` line already satisfies. Delete `/home/node/.codex` from the `mkdir -p` at `Dockerfile:114` and the test still passes, but the `jinn-codex` mount point is then root-owned, `node` cannot write `auth.json`, and the feature is dead in the container. Assert the mkdir line specifically.

---

## Checked and clean, so you can skip these

Recording the negative results so you don't spend time re-deriving them:

- **"Fully optional" holds.** Traced every path in `index.ts`: with `telegramAuth` absent, `!this.auth && !this.handler` collapses to the original guard, `ignoreOldMessagesOnBoot` keeps its relative position, `messageTextFromTelegram` reproduces `text || caption || ""` exactly, and `start`/`stop`/`replyMessage` short-circuit. Your `telegram-auth-optional.test.ts` covers it.
- **The 0.33.0 migration bundle is valid.** All four hashes recomputed and matched, base payload equals what v0.32.0 actually shipped, `MIGRATION.md` matches `manifest.json`, and `node scripts/instance-migration-bundle.mjs check --base-ref v0.32.0 --version 0.33.0` reports current.
- **No ReDoS in the new patterns.** All linear at 1,000,000-char adversarial inputs. `DISCOVERY_CODE_PATTERN` is quadratic on whitespace runs, but its only input is the 4096-byte-capped `discoveryTail` written by a local trusted process, so it caps out around 11 ms and is not remotely reachable. Worth flattening `\s*[:=]?\s*` to `\s*(?:[:=]\s*)?` eventually, not a blocker.
- **No `/g` `lastIndex` statefulness bug.** `ANSI_PATTERN` is only used with `replace`, `DISCOVERY_CODE_PATTERN` only with `matchAll`. Neither reaches `.test()`/`.exec()`.
- **`appendUtf8Tail` is correct**, including a forced cut inside a 4-byte character. No U+FFFD across 900 appends of multibyte text.
- **ANSI interleaving is handled**, including an escape sequence split across a chunk boundary.
- **`parseClaudeCallbackCode`'s URL accept set is tight.** Correctly rejects `https:`, `evil.com`, `localhost@evil.com` userinfo, `0.0.0.0`, `127.0.0.2`, `[::ffff:127.0.0.1]`, `/callback/`, `/Callback`. No SSRF surface since nothing is fetched.
- **No argv or shell injection.** `PROVIDERS[*].login` are frozen literals and `runCommand` uses `execFile`, never a shell.
- **Generation/stale-verification handling is correct.** The `generations` bump plus the `isCurrentGeneration` re-check properly drops a superseded status result; reproduced the worst interleaving and it behaves.
- **Timers and subscriptions are disposed correctly** on every path through `detach`; `active`/`pending`/`generations` are bounded by owners × 2.
- **Gates pass**: footguns 0 violations, ratchet clean, 70 telegram tests green.

Latent, no wrong outcome today: `auth-providers.ts:41` reads `failure.status`, but `execFile` errors carry `code`, so `exitCode` is always `1` on failure. Both providers only compare against `0`, so nothing misbehaves yet.

---

Findings 1 and 2 are the ones that decide whether this works at all. 3 is the one that affects users who never enable the feature. Happy to look again once those three are in.

hristo2612:
Verified `d5db3d3c` by running it, not by reading the diff. Four of the five findings are genuinely fixed, and the fixes are the right ones rather than the narrowest ones that would satisfy my repros.

## Confirmed fixed

**1. The latch.** Swept every chunk boundary in a full provider output line, with codex's real `Enter this one-time code:` phrasing:

```
boundaries=164   wrongURL=0 (truncated:0)   wrongCODE=0 (wrong-value:0)
```

Was 90/99 wrong on the URL. The `(?=\s)` terminator requirement, `updateDiscovery` superseding and re-arming `urlSent`/`codeSent`, and the `final` flush on exit together solve it properly. The other three symptoms too:

```
banner word : {"code":"WDJB-MJHT"}                                    (was "WARNING")
code= query : {"url":"...?code=A1B2C3D4E5&state=xyz"}  no code        (was code=A1B2C3D4E5)
banner link : {"url":"https://claude.ai/oauth/authorize?x=1"}         (was the docs URL)
```

Stripping URLs out of `codeSource` before matching is a nicer fix for the query-string case than the anchoring I suggested.

**2. `CODE_PATTERN`.** `/^[A-Z0-9]{4}-[A-Z0-9]{4,5}$/` rejects every token I listed:

```
DONE:-  TODO:-  ASAP:-  BLOCKED:-  LGTM:-  SHIP-IT:-  README:-  PLA-269:-
ICI-548:-  JIN-42:-  2026-08-27:-      WDJB-MJHT:INPUT  FHI3-TCJKF:INPUT
```

The `hasActiveFlow` scoping and the `index.ts` reordering are both right, and `scrubExplicitPayload` is exactly the distinction I was after: an unauthorized user's explicit `/auth … <secret>` still gets scrubbed, a bare token no longer does.

**3. `format.ts`.** All four cases, plus the `**bold**` case you added:

```
"**Release notes: https://…/v0.33.0**"     -> "Release notes: https://…/v0.33.0"
"*x https://a.com* and *y https://b.com*"  -> "x https://a.com and y https://b.com"
"see https://example.com/path_with_underscores ok"  -> unchanged
```

Bounds check on the placeholder restore is in.

**5. `stop()`.** Latched, and checked at all three points including after the `await` with cleanup.

**Housekeeping:** version back to 0.32.0, empty-owner set now refuses to construct and logs, and the Dockerfile assertion now pins the `mkdir` line. 68 telegram tests pass, CI green on 12.

---

## One blocker left: finding 4 is inverted, not fixed

The early return added at `auth.ts:86` runs **before** `sensitive` is computed, so `isCallbackUrlShape` never gets consulted when `parseAuthInput` succeeds:

```ts
if (!command && input && !this.hasActiveFlow(id)) return false;   // returns before...
const sensitive = Boolean(input) || AUTH_PAYLOAD_PATTERN.test(raw) || isCallbackUrlShape(raw);
```

Driven against the real `TelegramAuth`, owner id, private chat, no active flow:

```
PASSED-THRU  deleted=no    VALID callback URL, no active flow
CONSUMED     deleted=YES   INVALID callback URL, no active flow
```

So a callback URL carrying a **live** authorization code is passed through to the agent and left in Telegram history, while a malformed one is correctly scrubbed. The protection landed on the harmless case and the real secret is the one that leaks. Widening `CLAUDE_CALLBACK_CODE_PATTERN` to accept `_-` made this *more* reachable, since more real URLs now parse and take the early return.

Reachable whenever the flow is not active at paste time: after the 10 minute TTL fires, after `/auth_cancel`, or if the CLI exits once it has handed off to the browser.

**Fix:** compute `sensitive` and run `scrub` before the `hasActiveFlow` early return, or exclude callback-shaped input from that return. One line either way.

---

## Worth a second look, not a blocker

**The device-code keyword list is an allowlist with no fallback.** Removing the bare-line fallback was right, but the replacement is a fixed set of phrasings:

```
OK    "Enter this one-time code: WDJB-MJHT"      <- codex's actual output
OK    "Your device code is: …"  "User code: …"  "Verification code: …"
MISS  "Code: WDJB-MJHT"
MISS  "Enter the code WDJB-MJHT"
MISS  "Enter the following code in your browser: WDJB-MJHT"
```

It covers what codex emits today, so this is not breaking anything now, and the failure mode is soft (no code delivered, flow times out) rather than the wrong code. But a provider reworing its prompt silently breaks the feature with no test to catch it.

Worth reconsidering a context-free fallback: the reason the old bare-line fallback was dangerous was the loose `CODE_PATTERN`, and that is gone. `[A-Z0-9]{4}-[A-Z0-9]{4,5}` standing alone on a line is specific enough to be safe now, and it would make the extractor robust to wording instead of coupled to it.

**The CI gate change is a maintainer call, not yours.** Reverting the version bump broke `migration-bundle.yml`, and `--allow-unreleased` is a reasonable resolution: it is explicit, named, and `publish-npm.yml:121` still runs the strict check, so the release gate is intact. I am not objecting, but loosening a guardrail is my decision to make rather than something to land inside a feature PR. Leave it in and I will decide separately. Flag changes like that in the description next time instead of folding them into the diff.

(Latent, no impact: `isNewerVersion` compares `candidate.split(".")` positionally without normalising lengths, so a two-part version against a three-part one would compare against `undefined`. Every version here is three-part.)

---

**Verdict: not mergeable today, one line away.** Fix the callback ordering and I will merge. Genuinely good iteration on this round, the fixes went at the root cause rather than at my test cases.

PR #145, opened 2026-08-27
fix(web): reserve the top safe-area inset on the mobile chat list

theNailz:
> **Rebased onto `c3a91c18`.** The branch was cut from `12c1b976` and was 126 commits behind. Every `file:line` below is re-verified against the new base, and the two precedent citations that no longer exist there have been removed — see *The change*. Thanks for the review; it was right, and rebasing is what made it checkable.

## The bug

On an installed PWA on a Dynamic Island iPhone, the chat list's control band renders **under the iOS status bar**. The Focused/All filters, New chat and Search are visible but will not take a tap, because iOS keeps the touches in that strip.

Confirmed on device (1290x2796, so a 59px top inset). All four affected controls come from the same flex row:

| Where it lands on screen | Control | Line |
|---|---|---|
| Pill left of the clock | `focused` — half of the segmented control | `chat-sidebar.tsx:1549` |
| Over the clock digits | `all` — the other half of that same control | `:1550` |
| Across the wifi glyph | New chat | `:1580` |
| Right of the battery | Search chats | `:1589` |

## Why nothing above it clears the notch

| `file:line` | What it is | Provides a top inset? |
|---|---|---|
| `packages/web/src/routes/chat/page.tsx:810` | The chat route renders `<PageLayout chromeless>` | No — `chromeless` |
| `packages/web/src/components/page-layout.tsx:162` | `!chromeless && "pt-[var(--safe-top)] lg:pt-0"` | No, and this is exactly the branch being skipped |
| `packages/web/src/routes/chat/page.tsx:870` | Wrapper around the mobile `ChatSidebar` | No padding |
| `packages/web/src/routes/chat/page.tsx:858-859` | `ChatPageHeader` with `hideOnMobile={onMobileList}` | Hidden while the list is on screen |
| `packages/web/src/components/chat/chat-sidebar.tsx:1529` | The control band itself | **No — the gap** |

Three metas put the viewport behind the status bar in the first place: `index.html:9` (`viewport-fit=cover`), `index.html:16` (`black-translucent`) and `manifest.webmanifest:8` (`"display": "standalone"`).

## The change

Rung 1 of the Footprint Ladder — the `--safe-*` tokens already exist on `:root` (`globals.css:302-305`), with a comment prescribing exactly this composition at the call site. The band just was not following it:

```diff
-  "shrink-0 bg-[var(--sidebar-bg)] px-3 py-2 transition-shadow duration-150",
+  "shrink-0 bg-[var(--sidebar-bg)] px-3 pb-2 pt-[max(var(--safe-top),8px)] transition-shadow duration-150",
```

### Precedent

`grep -rn -- '--safe-top' packages/web/src | grep -v motion-safe` is the whole list. Every row below is one grep to check:

| `file:line` | What is already there |
|---|---|
| `packages/web/src/components/chat/chat-tabs.tsx:153` | `paddingTop: 'max(var(--safe-top), 0px)'` |
| `packages/web/src/routes/chat/session-picker.tsx:112` | `pt-[calc(var(--safe-top)+var(--space-12)+var(--space-2))]` |
| `packages/web/src/routes/todos/task-page/crumb-bar.tsx:67` | `pt-[calc(10px+var(--safe-top,0px))]` |
| `packages/web/src/plugins/sdk/plugin-notices.tsx:71` | `top-[max(var(--safe-top),var(--space-3))]` |
| `packages/web/src/components/auth/pairing-screen.tsx:54` | `py-[max(var(--safe-top),var(--space-8))]` |

The first two are the load-bearing ones: `chat-tabs.tsx` is the mobile thread header one component over, and `session-picker.tsx` is on this same route. The list screen is the only top-of-screen surface that did not follow the recipe.

**Two citations from the original description are gone and are not replaced.** Both were correct against the old base and are wrong against this one, so they are withdrawn rather than re-pointed:

- `pill-nav.tsx:400`, `:438` — deleted in `d8179b97` ("delete the write-only breadcrumb store and the unmounted PillNav chrome"). Worth stating plainly: `PillNav` had not been rendered by anything since `PageLayout` moved to the `NavRibbon` rail, so even on the old base this was a citation to **unmounted dead code**. It should not have been offered as precedent then either.
- `board-page.tsx:543` — that header was refactored into `PageScaffold`, and `page-scaffold.tsx` carries only `max(var(--safe-bottom), 6px)` (`:51`). The top inset was dropped in that refactor, so the precedent no longer exists.

### Why not `--chat-top-clearance`

It exists (`globals.css:326`) and looks like it fits, but it is `max(var(--safe-top), 12px)` and is scoped to clearance under the floating header pills (`chat/page.tsx:988`). Reusing it would have quietly moved the band's resting spacing from 8px to 12px. Composing `max(...,8px)` at the call site preserves `py-2` exactly.

### Behaviour

`max(...)` collapses to the original 8px wherever the inset is 0, so every desktop browser keeps its previous `py-2` spacing. The band is `shrink-0` in a `flex h-full flex-col`, so the widgets slot and the list scroller shift down with it; the scroller is `flex-1 min-h-0 overflow-y-auto` and simply gets shorter. Nothing at the bottom of the screen is touched.

Why a one-line fix shows as 9 added / 9 removed in `chat-sidebar.tsx`: one line is the class, one is the `data-` attribute, and the remaining seven are two adjacent comment blocks — both describing this same row, one of them stale — merged into one. That consolidation is not part of the fix; it is there because leaving two overlapping descriptions of a row whose behaviour just changed would be worse than merging them.

## Test

jsdom does no layout, so a reachability test passes with the bug in place. The guard asserts that the band consumes the inset **and** that the previously unreachable controls live inside that band, so it cannot be satisfied by padding an unrelated wrapper. `data-chat-list-controls` follows the existing convention (`data-chat-list-scroll`, `data-chat-mobile-header`).

Round trip re-run **on the rebased base**, with the marker left in place so the red exercises the assertion rather than the locator:

```
# padding reverted
AssertionError: expected '' to contain 'var(--safe-top)'
 Tests  2 failed | 1 passed (3)

# padding restored
 Tests  3 passed (3)
```

## Gates

Re-run after the rebase:

| Gate | Result |
|---|---|
| `pnpm test` (`packages/web`, full suite) | green — `399 passed (399)` files, `3439 passed (3439)` tests |
| `pnpm footguns` | green — `0 violations in 2 files, 0 suppressions` |

`typecheck` fails in my container with 15 `TS7006`/`TS2345` errors, all under `src/routes/workflow/` — and it fails **identically on a clean `origin/main` worktree with no diff applied**, so it isolates to a dependency my container does not build rather than to this change. Same story as before for `packages/jinn` (no prebuilt `pty.node`, no C toolchain) and `jinn-cli` lint (OOM-killed). CI is the authority on all three; this diff is confined to `packages/web/src/components/chat/`.

## Device verification

Unchanged by the rebase — measured on the original build, and the diff is byte-identical. Standalone on a 1290x2796 device: the Focused/All pill sits at **10.7 CSS px** before and **61.7 CSS px** after, a 51.0px shift = 59 (`env(safe-area-inset-top)`) − 8 (the previous `py-2`). In Safari, where the inset is 0, it stays at 10.7px — unchanged from before, which is the zero-inset case for the `max()` floor. Full measurement in [the earlier comment](https://github.com/hristo2612/jinn/pull/145#issuecomment-5437360926).

## Not covered here

**`NavRibbon` alignment at `lg` on an installed device.** The change is not gated on `variant`, which is intended and the test covers both. The consequence is on the desktop branch: the rail's `<nav>` (`packages/web/src/components/pill-nav.tsx:193-194`) is `px-1.5 pb-2.5 pt-3.5` and reserves no top inset — `pill-nav.tsx` now contains no safe-area handling at all. The rail is `hidden … lg:block`, so wherever the inset is non-zero at that width — an installed iPad PWA in landscape — the list band now drops by the inset and the rail beside it does not, and their top controls stop lining up. Net still an improvement, since previously both sat under the status bar and now one is clear. Out of this diff on purpose.

**The bottom tab bar.** Its padding computes to ~93 CSS px against a 34px home indicator on the same device, leaving a dead strip under the icons. Measured, different cause, tracked separately — deliberately out of this diff.


theNailz:
Verified on hardware, since the fix only does anything where the inset is non-zero.

Device: iPhone, 1290x2796 @3x (430x932 CSS px), installed to the home screen so the viewport runs behind the status bar. Measured by decoding the screenshots and profiling pixel rows.

| Build / context | Top of the Focused/All pill, from the screen top |
|---|---|
| before, standalone | **10.7 CSS px** — under the status bar |
| after, standalone | **61.7 CSS px** — clear of it |
| after, Safari (inset 0) | **10.7 CSS px** page-relative — unchanged from before |

The 51.0px shift is exactly 59 (`env(safe-area-inset-top)`) minus 8 (the previous `py-2`), which is what `pt-[max(var(--safe-top),8px)]` should produce. The Safari row is the zero-inset case: identical to the old build, so the `max()` floor leaves every non-notched surface untouched.

Confirming the standalone shots really were standalone: the top 59px paints `--sidebar-bg` (the band covering its own padding) rather than the `theme-color` Safari puts behind the status bar.

Tapping confirmed separately by the reporter — the controls were always visible, they just would not take a tap, so position alone would not have settled it.

Bottom tab bar measured identical before and after (106.3 CSS px below the icons), i.e. untouched by this diff. That bar reserving ~93px against a 34px home indicator is a separate issue, tracked separately and deliberately out of scope here.

theNailz:
## Before vs after

image

hristo2612:
Nice diagnosis, and the fix is the right one. Rung 1, one line, composed at the call site exactly as the token comment at `globals.css:301` prescribes. I verified the change itself rather than the description, and the code holds up.

**What I confirmed:**

- The diagnosis is accurate. `PageLayout chromeless` at `chat/page.tsx:810` is exact, the skipped branch is real, and `--safe-top` is defined at `globals.css:302`.
- Desktop is genuinely unaffected. `env(safe-area-inset-top)` is `0` off-device, so `max(0px, 8px)` collapses to the original `py-2`. No double-padding either: I traced the desktop tree from `PageLayout` down through `flex overflow-hidden h-full` and the `lg:flex` ribbon wrapper, and nothing above the band contributes a top inset.
- **You were right not to reuse `--chat-top-clearance`.** It exists at `globals.css:326` and looks like it fits, but it is `max(var(--safe-top), 12px)` and scoped to clearance under the floating header pills (`chat/page.tsx:988`). Reusing it would have quietly moved the band's resting spacing from 8px to 12px. Composing your own `max(...,8px)` preserves `py-2` exactly. Worth saying out loud so the next reviewer doesn't ask.
- **The test is genuinely non-vacuous**, and I checked that by hand rather than taking the round trip on faith. Reverting the padding and keeping the marker:

  ```
  AssertionError: expected '' to contain 'var(--safe-top)'
   Tests  2 failed | 1 passed (3)
  ```

  Byte for byte what you reported. Asserting `max(` with the reasoning in a comment, accepting either the class or the inline spelling, and pinning the four controls *inside* the marked band are all the right calls. That second test is what stops this from being satisfied by padding an unrelated wrapper.

Two things before merge, neither of them in the code.

---

## 1. Half the precedent table does not exist

This is the one that matters, because the PR is written as an evidence table and I would have merged on the strength of it.

| Claim | Reality |
|---|---|
| `pill-nav.tsx:400`, `:438` | The file is **266 lines**. It contains **no safe-area handling at all** — the only `safe` matches are `motion-safe:` utility classes. |
| `board-page.tsx:543` | Wrong path (`routes/todos/board/board-page.tsx`), and that file also has **no safe-area handling**, again only `motion-safe:`. |
| `chat-tabs.tsx:153` | Correct and exact. |
| `crumb-bar.tsx:67` | Correct and exact. |

Minor, same category: `page-layout.tsx:165` is actually `:162`.

The two real citations carry the argument fine on their own, so the conclusion stands. But please correct the table. Citations that pattern-match `motion-safe:` into a safe-area claim are worse than no citations, because they invite exactly the merge-on-trust I nearly did. If a precedent was inferred rather than read, leave it out.

## 2. `NavRibbon` will now be visibly misaligned on an installed iPad

`ChatSidebar` renders in two places and your change is not gated on `variant`, which is correct and your test covers both. The consequence is on the desktop branch: at `lg` on an installed iPad PWA, `env(safe-area-inset-top)` is non-zero, so the list band drops by the inset while `NavRibbon` sitting beside it does not.

`NavRibbon` is defined at `pill-nav.tsx:158`, in the very file cited above as already handling this. It does not, so the ribbon's top control stays under the status bar and now sits misaligned against the band next to it.

Net this is still an improvement, since previously both were unreachable and now one is fixed. Not a blocker and not something I want in this diff. Worth an issue, or a line under "Not covered here" alongside the bottom tab bar note.

---

Housekeeping on the diff itself: the 9 changed lines are 1 line of class, 1 `data-` attribute and a merge of two overlapping comment blocks that both described the same row. That consolidation is an improvement, not churn, and it is worth a sentence in the description so a reviewer does not have to work out why a one-line fix shows as 9/9.

CI is green across all 11 checks including Windows. Your note about the two legs you could not run locally is accurate: `packages/jinn` imports nothing from `@jinn/web` and this diff is confined to `components/chat/`, so CI covering them is the right call.

Fix the citation table and I am happy to merge this.

theNailz:
Rebased onto `c3a91c18` and rewrote the description. You were right on both points. Why we disagreed on the first one is worth a sentence, because it is the actual defect in my process.

## The citation table

The branch was cut from `12c1b976` and was **126 commits behind**. At that base `pill-nav.tsx` is 444 lines and does carry `top-[max(var(--safe-top),12px)]` at `:400` and `:438`, and `board-page.tsx:543` is the `--safe-top` header. Against current `main` — the tree you read — both are gone. `page-layout.tsx` is the same story: `:165` at the old base, `:162` at yours, and yours is the one that matters.

So the table was accurate when written and stale when reviewed. That distinction is worth nothing to you as a reviewer: a citation you cannot check is a citation you cannot check, and "it was true 126 commits ago" is not a defence. The fix was to rebase, not to argue.

One thing I found while re-checking that makes your point stronger rather than weaker. `pill-nav.tsx:400`/`:438` were deleted in `d8179b97`, whose message reads: *"nothing has rendered PillNav since PageLayout moved to the NavRibbon icon rail."* So even at the old base that was a citation to **unmounted dead code**. It should not have been offered as precedent then either, and your instinct about the table was catching something real that predates the staleness. `board-page.tsx`'s header went into `PageScaffold`, which carries only `max(var(--safe-bottom), 6px)` (`page-scaffold.tsx:51`) — the top inset was dropped in that refactor.

Both citations are **withdrawn rather than re-pointed**, and the description says so explicitly. The precedent table now contains only what `grep -rn -- '--safe-top' packages/web/src | grep -v motion-safe` returns on this base — five rows, each one grep to verify. The two load-bearing ones are `chat-tabs.tsx:153` (the mobile thread header, one component over) and `session-picker.tsx:112` (same route).

## NavRibbon

Confirmed, and on current `main` it is unambiguous: `pill-nav.tsx` contains **no** safe-area handling at all. The rail's `<nav>` is `:193-194`, `px-1.5 pb-2.5 pt-3.5`. Written up under "Not covered here" as you suggested, with the `lg`-only scope spelled out — the misalignment needs a non-zero inset at ≥1024px, i.e. an installed iPad in landscape.

## Housekeeping

The 9-added/9-removed explanation is in the description now: one line of class, one `data-` attribute, and seven lines of two overlapping comment blocks merged into one.

## Gates, re-run after the rebase

| Gate | Result |
|---|---|
| `packages/web` full suite | green — `399 passed (399)` files, `3439 passed (3439)` tests |
| `pnpm footguns` | green — `0 violations in 2 files, 0 suppressions` |
| red/green round trip | reproduced byte-for-byte on the new base |

`typecheck` fails in my container with 15 `TS7006`/`TS2345` errors, all under `src/routes/workflow/`. It fails **identically on a clean `origin/main` worktree with no diff applied**, so it is a dependency my container does not build rather than anything in this change. Flagging it rather than quietly reporting green. CI is the authority.

The fix itself is untouched — the rebase was clean and the diff is byte-identical to what you reviewed.

theNailz:
Heads-up on the two red checks now showing (`unit-tests (ubuntu-latest)`, `coverage`) — I do not think they are this diff, and here is the evidence rather than the assertion.

**The base is red.** `c3a91c18`, the commit this branch is now rebased onto, fails its own CI run ([33156777906](https://github.com/hristo2612/jinn/actions/runs/33156777906)) on the same jobs — `coverage`, `unit-tests (ubuntu-latest)` and `unit-tests (windows-latest)`. This PR inherited that.

**The failures are in a package this diff cannot reach.** Every failing test is in `packages/jinn`:

- `src/gateway/__tests__/todo-recovery.test.ts`
- `src/workflows/__tests__/todo-runs.test.ts`
- `src/sessions/__tests__/interrupted-message-carry-forward.test.ts`

The diff is confined to `packages/web/src/components/chat/`, and `packages/jinn` imports nothing from `@jinn/web`.

**They are non-deterministic.** Running the two `packages/jinn` files locally, alternating between a clean `origin/main` worktree and this branch to control for machine load:

| Round | clean `origin/main` | this branch |
|---|---|---|
| 1 | pass | FAIL |
| 2 | pass | FAIL |
| 3 | **FAIL** | **pass** |

Same command, same tree, different outcome — and round 3 inverts, which is what rules out a deterministic cause on either side. The assertion is `AssertionError: expected [] to have a length of 1 but got +0`, consistent with a race in these recovery/ledger tests.

One caveat I will state rather than hide: across all my local runs this branch failed more often than clean main (7/10 vs 2/7). I can offer no mechanism for that — a Tailwind class and a `data-` attribute in a React component cannot affect gateway recovery tests — and my harness symlinks a shared `node_modules` between the two worktrees, so cross-run interference is a live confound. I would not read the rate difference as signal, but I am not going to pretend I did not see it.

Not asking anyone to wave the red through — just isolating it from the change under review. If you want it green before merging, the base needs fixing first, or the run needs a retry.

hristo2612:
You are right and I was wrong. I checked your rebuttal against the base rather than taking it on trust, and every citation holds:

```
$ git show 12c1b976:packages/web/src/components/pill-nav.tsx | wc -l
444
$ ... | sed -n '400p;438p'
<div className="… left-[max(var(--safe-left),12px)] top-[max(var(--safe-top),12px)] …">
<div className="… right-[max(var(--safe-right),12px)] top-[max(var(--safe-top),12px)] …">

$ git show 12c1b976:packages/web/src/routes/todos/board/board-page.tsx | sed -n '543p'
<header className="flex-none px-5 pt-6 max-[700px]:pt-[calc(24px+var(--safe-top,0px))] …">

$ git show 12c1b976:packages/web/src/components/page-layout.tsx | sed -n '165p'
!chromeless && "pt-[var(--safe-top)] lg:pt-0",
```

444 lines, `:400` and `:438` both carrying `top-[max(var(--safe-top),12px)]`, `:543` the `--safe-top` header, `:165` the exact branch. Four for four. The only real error was a path segment, `todos/board-page.tsx` for `todos/board/board-page.tsx`, which is a typo and not what I accused you of.

I called that table fabricated. It was not. It was accurate at the commit you cut from, and I read it against a `main` that had moved 126 commits underneath you. Reviewing citations against the tip instead of the merge-base is my process defect, not yours, and "pattern-matched off `motion-safe:`" was an unfair characterisation of work you had actually done. Apologies for that.

I have adjusted how I review these: line-anchored claims get checked at the PR's merge-base first, and only then against the tip to see whether they have gone stale.

**On the red checks: your evidence is right too.** `c3a91c18` is the current `origin/main` tip and it fails its own CI run on `coverage`, `unit-tests (ubuntu-latest)` and `unit-tests (windows-latest)`. Your branch inherited exactly those three, plus the `all-checks-pass` roll-up. That is ours to fix, not yours, and it is not a merge blocker for this PR. I am sorting it out separately.

The code was already good on the first pass and nothing about it has changed. Standing by to merge once main is green underneath you.

PR #147, opened 2026-08-29
Feat/remote ssh employee

harrytwigg:
> **Draft — not ready to merge.** Opening it early to get your read on the approach
> before I polish it. Known outstanding items are listed at the bottom.

Hey @hristo2612, I'm a big fan of what you're doing here, so I thought I'd contribute
something back that I've been working on.

Before anything else: the thing I like most about Jinn is that it drives the real Claude
Code CLI inside a PTY, so work runs on a subscription rather than burning metered API
credits the way API-key-based agent platforms do. That was the deciding factor for me, and
it's a constraint I've been careful to preserve here — the remote environment strips
`ANTHROPIC_API_KEY` / `ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_BASE_URL` so an inherited key on
the other machine can't silently flip a session to metered billing, and nothing anywhere
falls back to `claude -p`.

## The problem

I run Jinn on a low-powered Raspberry Pi. It's a great always-on orchestrator, but there's
not much real work you can do on it — no serious builds, no test suites, and I don't want
repositories cloned onto it at all. Meanwhile I have a desktop sitting idle that could do
all of that comfortably.

So this PR adds the ability to move an employee's runtime off the gateway host and onto
another machine over SSH, **per employee**, while the gateway stays the orchestrator.

## What it adds

The current runtime wraps the Claude Code CLI locally. This adds an **alternative runtime**
alongside it: same engine, same PTY contract, but the process it owns is an `ssh` client
that launches Claude Code on a remote host instead.

It also adds a **custom command hook in the agent lifecycle** — a configurable point where
you can run whatever your setup needs before a session starts. In my case that's a script on
the Pi that talks to a PiKVM and physically powers my desktop on, then the chat simply
carries on once it's up. That turned out to be my favourite part: the employee is just
*there*, and the machine being off is no longer something I have to think about.

Three optional fields on an employee, threaded like `cliFlags` already is:

```yaml
remoteHost: build-box
remoteUser: jinn
remoteCwd: /srv/jinn-work/main
```

When `remoteHost` is set, `InteractiveClaudeEngine` spawns `ssh` instead of `claude`, and
the node-pty PTY it owns becomes the local ssh client's — which carries the remote TUI.
That substitution is most of the change: `node-pty` was never Claude-specific, so
scrollback, resize, kill and the safety-prompt parser all keep working untouched.

The design goal was that a remote employee is a **full** employee, not a degraded one:

| | Where it runs |
|---|---|
| Todos, Notes, work items, delegation, knowledge | Gateway, via the built-in `jinn` MCP server over the tunnel |
| Skills, `docs/`, `knowledge/`, `org/` | Gateway, via a mounted instance home |
| Hooks — turn completion, tool events, safety prompts | Delivered to the gateway over a reverse tunnel |
| The dashboard's live terminal | Attached to the remote PTY |
| Repository code, builds, tests | The remote host — the point of the change |

## The parts that needed care

**Hook delivery without touching the auth path.** `validateHookPost` rejects any
non-loopback address before it even checks the secret, and I didn't want to weaken that.
Instead the session's `ssh` carries a reverse forward, so the remote `hook-relay.mjs` posts
to `127.0.0.1:<forwarded port>` and the request arrives as genuine loopback traffic. The
`isLoopback` check is completely unchanged. The port is probed on the remote immediately
before the spawn and paired with `ExitOnForwardFailure=yes`, so losing that race is a fast,
visible exit rather than a session whose hooks can never arrive.

**MCP.** Every server spec carries gateway-local absolute paths — not just the builtin, but
every third-party one, because `wrapServersWithScrub` rewrites them all to
`process.execPath` + the gateway's `dist`. There's a pure `remapMcpConfigForRemote` for
this, and the remote host needs a matching `jinn-cli` install (never started as a daemon —
it only supplies the entrypoints). A version mismatch is refused at spawn rather than
surfacing later as a confusing MCP failure.

**Everything an employee reads and writes that isn't code.** The gateway's instance home is
sshfs-mounted on the remote host, and the session's `$JINN_HOME` is a symlink farm over it,
rebuilt every spawn — so `knowledge/`, `docs/`, `org/` and `skills/` are the gateway's real
files, not copies. `gateway.json` and `tmp/` are staged for real instead, since the former
must name the tunnel port.

**Three failure modes that are silent by nature**, each closed deliberately, because for a
"kick it off and walk away" employee a silent hang is the worst outcome available:

- Claude Code's **folder-trust dialog** doesn't match `parsePermissionPrompt`'s strict shape
  and fires no `Notification` hook, so the first turn against an unseeded directory would
  hang forever. A staged seeder pre-trusts it, sharing its merge logic with `seedTrust`.
- The **dashboard's idle PTY** had no employee context. Left alone it spawns `claude` on the
  gateway and — worse — that PTY is *adopted as the session's warm handle*, so the next real
  turn pastes into a local process while the UI looks entirely normal. The idle path goes
  over SSH too.
- A **rate limit** would relocate the turn: the same-engine retry rebuilds its opts from
  scratch, and substitution hands the work to an engine with no remote support. The retry
  now restates the target; substitution is suppressed for remote employees.

**A mount that silently unmounts** is the remaining divergence risk, so a sentinel is
verified before every spawn (never cached), and the existing PreToolUse policy gate is
extended to refuse instance-home writes landing outside the verified mount.

**Hosts that are usually off.** This is the lifecycle hook above, and a desktop isn't a
server, so waking it is part of the feature rather than an error path: `wakeCommand` (or a
built-in Wake-on-LAN packet for the simple case), the session moved to `waiting`, the
operator notified, then bounded polling. The command gets a generous budget on purpose —
mine spends most of a minute reading a power state over the network before it presses
anything, and killing it in that window would mean no press at all and a turn that times out
having done nothing. Two deliberate
asymmetries — the dashboard's idle path never wakes anything, and none of this touches the
engine-health store, which is keyed by engine *name* and would otherwise hold back every
local employee's turn because one desktop is asleep.

## Deliberately cut

- **Live SSE token streaming.** A remote session runs without the per-PTY proxy. Results
  still arrive in full on the `Stop` hook. This is what makes the change tractable — the
  proxy binds loopback on the gateway and is unreachable from another host.
- **Attachments**, refused for remote employees: `buildAttachmentSuffix` appends
  gateway-local paths that name nothing on the other machine.

## Known gaps, documented rather than hidden

- Cost and transcript accounting read empty for remote turns — the gateway reads transcripts
  from its own home, and the remote CLI writes them on the remote host. `--resume` is
  unaffected, since that CLI manages its own transcript consistently across turns.
- Lost-`Stop` recovery degrades for the same reason: a genuinely stalled remote turn fails
  rather than recovering its text.
- The gateway's API bearer and any MCP server API keys land in 0600 files on the remote
  host, and the mount makes the instance home readable there. That's inherent to the
  employee being as capable as a local one — the remote host is inside the same trust
  boundary as the gateway, and the docs say so plainly.

## Things I only found by running it

Worth mentioning because none of them were visible from the code:

- A **non-interactive `ssh` reads no rc file**, so a host whose Node comes from a version
  manager reports no `node` at all — and `nvm.sh` can't rescue that, because it's bash-only
  and `/bin/sh` is `dash` on Debian-family systems. Worse than the failed preflight: Claude
  Code invokes every hook as bare `node`, so no hook could ever start and every turn would
  hang. nvm's layout is now read directly, honouring its `default` alias (a global
  `jinn-cli` lives under one version's tree).
- **ssh consumes only the first `--` it sees.** With a guard `--` before the destination and
  another after it, the remote shell received `-- cd …` and answered
  `/bin/bash: --: invalid option` — every spawn died instantly. My own test asserted the
  malformed shape, so it stayed green throughout.
- **Profile managers.** A wrapper that sets `CLAUDE_CONFIG_DIR` also `unset`s every
  `CLAUDE_*` variable, stripping the ones a session needs. There's a `claudeConfigDir`
  setting instead, and the trust seed runs under the same profile — Claude Code keeps
  `.claude.json` *inside* that directory, so a seed under the wrong one is indistinguishable
  from no seed at all.

## Testing

Running on my own Pi (gateway) and desktop (remote), end to end: hooks arriving over the
tunnel, MCP tools answering, skills loading through the mount, `--resume` across turns.

Unit tests cover the pure surfaces — argv construction and shell quoting through a real
`sh`, the MCP path remap, containment, profile resolution, the wake budget. Where a claim is
of the "this hole is now closed" kind, I did the red round trip AGENTS.md §1 asks for.

## Outstanding before this is mergeable

Being upfront about what isn't done:

- [ ] **13 footgun violations** — 9 `spawn-without-stdio` and 3 literal `7777` in my tests,
      plus one `scattered-env` in `claude-interactive.ts`. CI would fail on these.
- [ ] One stale test assertion left over from the `--` fix.
- [ ] No new CI job, so nothing to add to `all-checks-pass` — but worth discussing whether
      this warrants one, since the interesting behaviour needs two machines.
- [ ] Windows leg unexercised. The remote side is POSIX by assumption; the gateway side
      should be fine, but I haven't run it.
- [ ] Happy to split this up if it's too much for one PR — the transport, the mount, and the
      wake support are fairly separable.

Very open to being told the whole approach is wrong, or that you'd rather solve this
differently.


harrytwigg:
I'm gonna close this now. Guessing it's not something you want at the moment. Let me know if you'd like something like this in the future, and I will reopen it.

Been working on some other great things that I think upstream could benefit from e.g. Open Code provider support

